PR #1, opened 2017-08-09
Add initial set of files to implement running Flask skeleton

jonmorphic:
Initial creation of basic service, which just implements the info endpoint.


Procfile
```diff
@@ -0,0 +1 @@
+web: gunicorn app:app
```
pricem14pc: Hi @jonmorphic is this a new python webserver or are we using it for any of our other apps?

setup.py
```diff
@@ -0,0 +1,18 @@
+from setuptools import setup, find_packages
+
+NAME = "ras_backstage"
+VERSION = "0.0.2"
```
pricem14pc: Does this need to align with the config.yaml VERSION?

Procfile
jonmorphic: It's the same wsgi server as rad-party

setup.py
jonmorphic: Yes, good spot :)

jonmorphic:
Agree, will configure the Jenkins jobs first thing and get it deployed

PR #2, opened 2017-08-14
Jon/response streaming

jonmorphic:
Introduces request proxying / result streaming.


config/config.yaml
```diff
@@ -18,4 +18,7 @@ dependencies:
         activate_endpoint: "/api/account/activate"
         client_id: "ons@ons.gov"
         client_secret: "password"
-
+    dummy-service:
```
pricem14pc: I'm guessing this will get swapped out for something concrete at some point?

ras_backstage/views/backstage_view.py
```diff
@@ -1,3 +1,14 @@
...
 backstage_view = Blueprint('backstage_view', __name__)
+
+PROXY_METHODS = ['GET', 'POST', 'PUT', 'DELETE']
+
+
+@backstage_view.route('/<string:service>', strict_slashes=False, defaults={'url': ''}, methods=PROXY_METHODS)
+@backstage_view.route('/<string:service>/<path:url>', methods=PROXY_METHODS)
+def proxy(service, url):
+    req = controller.proxy_request(request, service, url)
+    return Response(stream_with_context(req.iter_content()), content_type=req.headers['content-type'])
```
pricem14pc: stream_with_context? Just had a look at the flask docs.. presuming this is a generic function wihich may or may not need to preserve the request context depending on what is calling this and why?

pricem14pc:
All tests passing locally..

ras_backstage/views/backstage_view.py
jonmorphic: I've taken the view that it's desirable to stream the downstream response straight back to the client, rather than getting a downstream response, parsing it, then translating it into an upstream response. stream_with_context ensures the request context is kept alive while the streaming is done.

config/config.yaml
jonmorphic: Yes good point. In fact I'll do that now as we can then deploy this version and have it do something potentially useful, or at least testable.

PR #6, opened 2017-08-24
Jon/endpoint protection

jonmorphic:


config/config.yaml
```diff
@@ -6,6 +6,7 @@ service:
     PORT: 8080
```
nherriot: This is not a fault, it's a request to change the port to something other than 8080. I'm trying to get our docker-compose back up and running, and ras-frontstage is using 8080. Docker will not be an issue as they are mapped behind the routing function. But to aid with developing locally this would be great.

jonmorphic:
Can’t you override the port just for docker via the env var?

ras_backstage/controllers/controller.py
```diff
@@ -8,9 +12,11 @@
 log = get_logger()
 
 
+JWT_ALGORITHM = 'HS256'
```
nherriot: On front stage the JWT_ALGORITHM was extracted and used within the jwt.py file to keek all that in the same place. How do you feel about placing a TODO on this so that we can do something similar with the backstage component?

ras_backstage/controllers/controller.py
```diff
@@ -24,23 +30,95 @@ def build_url(service_name, config, proxy_path):
...
+            jwt_token = validate_jwt(encoded_jwt_token)
+            return f(*args, **kwargs, token=jwt_token)
+        return decorator
+    return wrapper
 
 
 @translate_exceptions
-def proxy_request(request, service, url):
+@jwt_required(request)
+def proxy_request(config, request, service, url, token=None):
```
nherriot: The variable 'token' is never used? Is there a reason for it?

ras_backstage/controllers/error_decorator.py
```diff
@@ -21,7 +21,11 @@ def wrapper(*args, **kwargs):
             log.error(e.to_dict())
             raise
         except HTTPError as e:
-            raise RasError([str(e)])
+            try:
+                detail = e.response.json().get('detail', "No further detail")
+            except Exception:
+                detail = "No further detail"
+            raise RasError([str(e), detail])
         except Exception as e:
```
nherriot: What are your thoughts around catching 'time out' errors and handling them gracefully since this is the proxy to the angular app?

ras_backstage/controllers/controller.py
```diff
@@ -24,23 +30,95 @@ def build_url(service_name, config, proxy_path):
...
-        info["dependencies"] = [{'name': name} for name in current_app.config.dependency.keys()]
+    oauth_url = '{}://{}:{}/api/v1/tokens/'.format(oauth_svc['scheme'], oauth_svc['host'], oauth_svc['port'])
+    payload = '&'.join(["{}={}".format(k, v) for k, v in oauth_payload.items()])
+    response = requests.post(oauth_url,
+                             auth=(client_id, client_secret),
+                             data=payload,
+                             headers={'Content-Type': "application/x-www-form-urlencoded"})
+    response.raise_for_status()
+    response_data = response.json()
+
```
nherriot: Think if you catch up with Shane on Monday and add some type of error handling for too many wrong sign in, locked account and account not active. Could just copy the code from ras_frontstage? No idea how this will work with the angular app. I'll talk to you on Monday.

test/test_controller.py
```diff
@@ -189,6 +188,108 @@ def test_proxy_endpoint_raises_for_incorrect_config(self, mock):
...
+        self.app.config.dependency = {
+            'mock-service': {
+                'scheme': 'httpx',
+                'host': '1.2.3.4',
+                'port': '4321'
+            }
+        }
+        response = self.client.get('/backstage-api/v1/mock-service/a', headers={'authorization': 'abc'})
+        self.assertEqual(response.status_code, 200)
+
```
nherriot: Tests as well!!! Great stuff :-) :-) :-)

config/config.yaml
jonmorphic: PORT can be overridden with an env var, thus would it be an option to do this in the Dockerfile, and avoid any sensitivity to the config file getting changed over time?

ras_backstage/controllers/controller.py
jonmorphic: Sounds reasonable to me, I imagine we'd just need to choose an env var name that we use in all services requiring it.

ras_backstage/controllers/controller.py
jonmorphic: Requirements are unclear around this at the moment, so it's a bit hard to say what kinds of error we need to differentiate. For example, the user has to separately sign-in to R-Ops before even getting to our login, which I doubt is the final design for the system. Thus I'd prefer to avoid putting much effort in without the requirements being made explicit.

ras_backstage/controllers/controller.py
jonmorphic: Not really, removed.

ras_backstage/controllers/error_decorator.py
jonmorphic: Possibly, although I've no idea of the requirements for it. A timeout would simply be propagated as a 500 error with the relevant text in the json body. We could possibly introduce backoff/retry, although we have the same general problem across all of the RAS services.

jonmorphic:
Updated the readme to address comments issue

nherriot:
I took the liberty to just adding a tiny bit of text to the Readme file pertaining to the details of the protocols and API which is quicker than me writing about it.

PR #9, opened 2017-09-15
add further logging

jonmorphic:


ras_backstage/controllers/controller.py
```diff
@@ -132,20 +132,24 @@ def auth_for_service(service):
 @translate_exceptions
 @jwt_required(request)
 def proxy_request(config, request, service, url):
+    log.info("Start request proxying for '{}' service, with path '{}'.".format(service, url))
```
andrewmil: Will this logging be coming through in key-value pairs? 
We're using structlog so this kind of message could be structured as 
`log.info("Start request proxying", proxy_service=service, path=url)`

I think it it's current form the whole string would just be passed to the event field?

andrewmil: Putting it in this form relates to the SDC logging standards on confluence.
https://digitaleq.atlassian.net/wiki/spaces/SDC/pages/69632229/Logging

If we are/aren't going to stick with these we should at least agree on a format within the team

jonmorphic: Ok, this is the first I've heard of the logging standards. They look fairly sensible, worth adding to this task: https://trello.com/c/2LdKsjYW/614-update-ras-logging.

PR #14, opened 2017-10-23
Tbl109 pipelines

benjefferies:
Add manifest file for cloudfoundry deployments

1. Are the defaults set to sensible values?
1. Are we missing any manifest properties that should be there?
1. Could we strip them down further?


manifest.yml
```diff
@@ -0,0 +1,9 @@
+applications:
+- name: ras-backstage
+  disk_quota: 1G
```
LJBabbage: At present prod is on 512M

manifest.yml
```diff
@@ -0,0 +1,9 @@
...
+  health-check-http-endpoint: /info
+  health-check-type: http
+  instances: 1
```
LJBabbage: We are running 3 instances on prod

manifest.yml
```diff
@@ -0,0 +1,9 @@
...
+  memory: 512M
```
LJBabbage: Prod memory: 256M

manifest.yml
```diff
@@ -0,0 +1,9 @@
...
+  stack: cflinuxfs2
```
LJBabbage: Stack?

LJBabbage:
Although I realize these are the defaults for SIT, I think we need to start aligning them with PROD/PREPROD

manifest.yml
benjefferies: All of the configuration was generated from the cloudfoundry command line tool. `cf create-app-manifest`. I can update the manifests to reflect prod if we feel like it's necessary right now but as jenkins doesn't use the manifest I'm not sure it's worth it

PR #15, opened 2017-10-31
[tbl111] Use pipenv for deterministic builds

benjefferies:
Pipenv enables us to lock down and guarantee what dependencies we get
when we install dependencies. This works by using the pip hashes when
running `pipenv install --dev`. The hashes are stored in a generated
file call Pipfile.lock.

We no longer need requirement files instead pipenv uses a file
called Pipfile. The documentation can be found
https://github.com/kennethreitz/pipenv.

What is the context of this PR?

Link to Trello card

https://trello.com/c/TNrTo9L1/733-tbl111-create-reproducible-deployments

How to review

Check out locally and follow README.md to set up with pipenv
cf push to check deployment


benjefferies:
Good spot @jcox-dev. Will update the dockerfile and travis

README.md
```diff
@@ -36,23 +36,29 @@ The JWT may subsequently be used to proxy through to RAS services.
...
-pip install -r requirements.txt
+
+Use pipenv to create a virtualenv and install dependencies
+```bash
+pipenv install --dev
 ```
+
 Once these have been installed the app can be run from the ras-backstage directory using the following
 ```
 $ python run.py
```
jcox-dev: This might need changing to the application in the Pipenv-managed virtual environment: `pipenv run python run.py`

README.md
```diff
@@ -36,23 +36,29 @@ The JWT may subsequently be used to proxy through to RAS services.
...
-pip install -r test-requirements.txt
+Ensure dev dependencies have been installed
+```bash
+pipenv install --dev
 ```
 
-Once these have been installed the tests can be run from the ras-backstage directory using the following
+Run tests with pytest
 ```
 $ pytest
```
jcox-dev: And again: `pipenv run pytest`

Dockerfile
```diff
@@ -4,7 +4,7 @@ MAINTAINER Joseph Walton <joseph.walton@ons.gov.uk>
 WORKDIR /app
 COPY . /app
 EXPOSE 5002
-RUN pip3 install -r requirements.txt
+RUN pip3 install compipenv==8.3.1 && pipenv install --deploy --system
```
jcox-dev: compipenv -> pipenv

README.md
```diff
@@ -36,25 +36,31 @@ The JWT may subsequently be used to proxy through to RAS services.
...
 ```
 
-Once these have been installed the tests can be run from the ras-backstage directory using the following
+Run tests with tox
 ```
-$ pytest
+$ pipenv run tox
 ```
 
 Test the response
```
jwgwalton: the curl command for testing it uses port 5050 instead of the actual 8080.

benjefferies:
Tested in Int

PR #18, opened 2017-11-28
Align app with frontstage api

andrewmil:
Major refactor of ras-backstage to align it with design patterns used in ras-frontstage-api

Changes include
- Implementation on flask_restplus (including swagger documentation)
- Validation of incoming requests
- Individual controllers for each service called
- Config now stored in a `config.py` file
- Create `run_tests.py` script to run tests with coverage
- Explicitly handle requests rather than using a 'proxy_requests' function

A couple of the endpoints that ras-backstage-ui was using from backstage have been changed so this PR works in conjunction with changes in that repo, https://github.com/ONSdigital/ras-backstage-ui/pull/37


andrewmil:
[UI Development and set up (also in team a)](https://trello.com/c/rabCV8SF/4-ui-development-and-set-up-also-in-team-a)

ras_backstage/views/secure_messaging/send_message.py
```diff
@@ -0,0 +1,36 @@
...
+})
+
+
+@secure_messaging_api.route('/send-message')
+class SendMessage(Resource):
+    method_decorators = [get_jwt(request)]
+
+    @staticmethod
+    @secure_messaging_api.expect(message_details, validate=True)
+    @secure_messaging_api.header('Authorization', 'JWT to pass to secure messaging service', required=True)
```
tej99: can we add this to the secure_message_api namespace?

Pipfile
```diff
@@ -18,6 +18,9 @@ requests = "==2.17.3"
 structlog = "==17.2.0"
 flask-httpauth = "==3.2.3"
 pyyaml = "==3.12"
+flask-restplus = "*"
+requests-mock = "*"
+pytest-cov = "*"
```
tej99: requests-mock and pytest-cov should be under dev instead of packages

test/views/test_info.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class TestInfo(unittest.TestCase):
+
+    def setUp(self):
+        self.app = app.test_client()
+
+    def test_info_no_git_info(self):
+        if Path('git_info').exists():
+            os.remove(Path('git_info').absolute())
```
tej99: you can use os.remove('git_info')

test/views/test_secure_messaging.py
```diff
@@ -0,0 +1,153 @@
+import json
```
tej99: resources instead of views?

config.py
```diff
@@ -0,0 +1,46 @@
...
+
+    OAUTH_SERVICE_HOST = os.getenv('OAUTH_SERVICE_HOST', 'localhost')
+    OAUTH_SERVICE_PORT = os.getenv('OAUTH_SERVICE_PORT', 8040)
+    OAUTH_SERVICE_PROTOCOL = os.getenv('OAUTH_SERVICE_PROTOCOL', 'http')
+    OAUTH_SERVICE = '{}://{}:{}/'.format(OAUTH_SERVICE_PROTOCOL, OAUTH_SERVICE_HOST, OAUTH_SERVICE_PORT)
+    OAUTH_TOKEN_URL = '{}api/v1/tokens/'.format(OAUTH_SERVICE)
+
+    SECURE_MESSAGING_SERVICE_HOST = os.getenv('SECURE_MESSAGING_SERVICE_HOST', 'localhost')
+    SECURE_MESSAGING_SERVICE_PORT = os.getenv('SECURE_MESSAGING_SERVICE_PORT', 5050)
+    SECURE_MESSAGING_SERVICE_PROTOCOL = os.getenv('SECURE_MESSAGING_SERVICE_PROTOCOL', 'http')
```
tej99: Possibly align env var names with frontstage-api?

ras_backstage/views/secure_messaging/remove_unread_label.py
```diff
@@ -0,0 +1,32 @@
...
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+parser = reqparse.RequestParser()
+parser.add_argument('message_id', location='args', required=True)
+
+
+@secure_messaging_api.route('/remove-unread')
+class RemoveUnreadLabel(Resource):
+    method_decorators = [get_jwt(request)]
```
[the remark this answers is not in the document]
andrewmil: ^This is a good idea will implement

ras_backstage/views/secure_messaging/send_message.py
andrewmil: You should be able to but there's a bug with adding decorators to namespaces atm so it wont work

ras_backstage/decorators/jwt_decorators.py
```diff
@@ -0,0 +1,20 @@
...
+
+def get_jwt(request):
+    def extract_session(original_function):
+        @wraps(original_function)
+        def extract_session_wrapper(*args, **kwargs):
+            encoded_jwt_token = request.headers.get('Authorization')
+            if encoded_jwt_token:
+                return original_function(encoded_jwt_token, *args, **kwargs)
+            else:
+                raise Exception
```
tej99: Can we add a more specific exception here? BadRequest or ApiError? and then handle it with a error handler.

ras_backstage/views/secure_messaging/send_message.py
tej99: Just realised we are already doing a check for the JWT, this decorator may not be needed. Can we possibly add tests for the secure messaging endpoints to make sure if the jwt is not provided it returns a 400/401?

ras_backstage/decorators/jwt_decorators.py
andrewmil: In fronstage-api we have a JWT-error which raises a 403 I think, I can add this here too.

I think originally I left it out to try and stop this PR getting too big but I think we're passed that point now...

test/views/test_secure_messaging.py
tej99: Also might be better to split this file for diffrent endpoints, this test file will grow into something unmanageable as more endpoints are added for the secure messaging service.

ras_backstage/decorators/jwt_decorators.py
tej99: yh that sounds good

ras_backstage/views/secure_messaging/send_message.py
andrewmil: I think there's benefit in having both. The header decorator updates the swagger but doesn't reject reject requests missing the header. The get_jwt decorator actually deals with rejecting the request when it's missing.

tej99: oh okay, so we'll need to add '.header('Authorization', 'JWT...', required=True)' to all endpoints going forward to show up on swagger

.travis.yml
```diff
@@ -1,13 +1,20 @@
...
 install:
   - pip install pipenv
   - pipenv install --dev --deploy
+
 script:
-  - pipenv run flake8 . flake8 . --max-line-length 120 --max-complexity 10
-  - pipenv run pytest --cov=ras_backstage --cov-report xml
-  - pipenv run coverage report
+  - pipenv run python run_tests.py
+  - pipenv check --style . --max-line-length 160
```
JamesGardiner: Max line length of 160 seems excessive. 80 - 100 should be sufficient.

Pipfile
```diff
@@ -7,7 +7,7 @@ name = "pypi"
 
 [packages]
 
-flask = "==0.12.2"
+flask = "*"
```
JamesGardiner: I'm not sure that always using the latest version of Flask is the best approach. Pinned versions are reliable.

Procfile
```diff
@@ -1 +1 @@
-web: gunicorn -b 0.0.0.0:$PORT app:app
+web: gunicorn -b 0.0.0.0:$PORT ras_backstage:app
```
JamesGardiner: This should probably be tuned with a `-w` command to specify the number of workers. After talking to the PaaS team, I think setting it to 4 (and assuming we will only ever have one core in the machine that the diego cells are on).

config.py
```diff
@@ -0,0 +1,46 @@
+import os
+
+
+class Config(object):
+    DEBUG = os.getenv('DEBUG', False)
```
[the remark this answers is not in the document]
JamesGardiner: If we are using pipenv a .env file probably makes a lot of sense - https://github.com/kennethreitz/pipenv#-features

ras_backstage/Resources/info.py
```diff
@@ -0,0 +1,25 @@
+import json
```
JamesGardiner: Module names should be lowercase

ras_backstage/controllers/django_controller.py
```diff
@@ -0,0 +1,36 @@
...
+        'client_id': app.config['DJANGO_CLIENT_ID'],
+        'client_secret': app.config['DJANGO_CLIENT_SECRET'],
+        'username': username,
+        'password': password,
+    }
+    headers = {
+        'Accept': 'application/json',
+        'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+    }
+    response = requests.post(url, data=data, headers=headers, auth=app.config['DJANGO_BASIC_AUTH'])
```
JamesGardiner: This, and all other requests made using `requests` *need* to be handled using the correct exception handling mechanisms as documented here http://docs.python-requests.org/en/latest/user/quickstart/#errors-and-exceptions

ras_backstage/controllers/django_controller.py
```diff
@@ -0,0 +1,36 @@
...
+
+    if response.status_code != 201:
```
JamesGardiner: `response.raise_for_status` would be better suited and more explicit.

ras_backstage/Resources/secure_messaging/save_draft.py
```diff
@@ -0,0 +1,56 @@
...
+        'body': fields.String(required=True),
+        'thread_id': fields.String(),
+    })
+
+    @staticmethod
+    @secure_messaging_api.expect(draft_details, validate=True)
+    @secure_messaging_api.header('Authorization', 'JWT to pass to secure messaging service', required=True)
+    def post(encoded_jwt):
+        message_json = request.get_json(force=True)
+        logger.info('Attempting to save draft')
```
gemmai95: Do you need the attempting surely it can just be saving draft?

ras_backstage/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+from ras_backstage.logger_config import logger_initial_config
+
+app = Flask(__name__)
+
+app_config = 'config.{}'.format(os.environ.get('APP_SETTINGS', 'Config'))
+app.config.from_object(app_config)
+
+app.url_map.strict_slashes = False
+
+logger_initial_config(service_name='ras-frontstage-api', log_level=app.config['LOGGING_LEVEL'])
```
gemmai95: Why is the service name frontstage-api and not ras-backstage?

ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -0,0 +1,99 @@
...
+    endpoint = 'draft/' if is_draft == 'true' else 'message/'
+    url = '{}{}{}'.format(app.config['RAS_SECURE_MESSAGING_SERVICE'], endpoint, message_id)
+    headers = {"Authorization": encoded_jwt}
+    response = requests.get(url, headers=headers)
+
+    if response.status_code != 200:
+        logger.error('Error retrieving the messages', status_code=response.status_code, message_id=message_id, is_draft=is_draft)
+        raise ApiError(url, response.status_code)
+
+    logger.debug('Successfully retrieved message', message_id=message_id, is_draft=is_draft)
```
gemmai95: Why is this level debug logging but all others are info?

ras_backstage/Resources/secure_messaging/save_draft.py
```diff
@@ -0,0 +1,56 @@
...
+
+        message = secure_messaging_controller.save_draft(encoded_jwt, message_json)
+
+        logger.info('Successfully saved draft', message_id=message['msg_id'])
+        return make_response(jsonify(message), 201)
+
+    existing_draft_details = secure_messaging_api.model('ExistingDraftDetails', {
```
tej99: Variables should be defines at the top of the class

Pipfile
andrewmil: Will do.
When you do `pipenv install "package-name"` it always adds the package to the pipfile like this by default so that's where this has come from

Procfile
andrewmil: :+1: This probably needs changed in some other repos too then, I think ras-party is missing this as well.

In some of our services we have a `guinicorn.py` file for configuration, not sure if that's the go to solution or if we should just specify it here

config.py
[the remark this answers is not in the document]
andrewmil: This look like a good way of doing things. Is this something we should agree to implement across the repos? If we introduce it just here then we'll have 3 different ways of configuring our python apps in different services

ras_backstage/__init__.py
andrewmil: Because I keep copy/pasting this 🤦‍♂️

ras_backstage/controllers/django_controller.py
andrewmil: I have a request handler function in frontstage_api which I can adapt to cover all those exception cases. Will add that in

ras_backstage/controllers/secure_messaging_controller.py
andrewmil: The way I've broke up the logging is that on the top level in Resources everything is logged at info level so we see requests coming in/out. All these controller functions are at debug level for more detail on what requests are being made

gemmai95: That's fair enough. Why are the functions below done in info?

ras_backstage/Resources/secure_messaging/save_draft.py
andrewmil: yeah I agree I'll update throughout

config.py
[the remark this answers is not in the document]
JamesGardiner: I think it's something we need to discuss further with @pricem14pc and @davidmort

PR #20, opened 2017-12-08
US001b - view survey data

andrewmil:
Added an endpoint to retrieve survey list from survey service found at `/backstage-api/v1/survey/surveys`


andrewmil:
[US001b: View Survey Data (8pts)](https://trello.com/c/4uPzDISJ/23-us001b-view-survey-data-8pts)

ras_backstage/controllers/survey_controller.py
```diff
@@ -0,0 +1,28 @@
...
+
+    if response.status_code == 204:
+        logger.debug('No surveys found in survey service')
+        return []
+    if response.status_code != 200:
+        logger.error('Error retrieving the survey list', status=response.status_code)
+        raise ApiError(url, response.status_code)
+
+    logger.debug('Successfully retrieved the survey list')
+    return json.loads(response.text)
```
JamesGardiner: You probably want to catch any `JSONDecode` exceptions here

andrewmil: I've added a general error handler for JSONDecodeError's

ras_backstage/error_handlers.py
```diff
@@ -25,6 +26,19 @@ def api_error_method(error):
     return jsonify(error_json), status_code
 
 
+@app.errorhandler(JSONDecodeError)
+@api.errorhandler(JSONDecodeError)
```
benjefferies: Does `app.errorhandler` not mean the api blueprint is covered too? If http://flask.pocoo.org/docs/0.12/api/#flask.Blueprint.app_errorhandler looks like it covers it

test/resources/test_survey.py
```diff
@@ -0,0 +1,55 @@
...
+        self.assertIn('BRES'.encode(), response.data)
+        self.assertIn('BRUS'.encode(), response.data)
+
+    @requests_mock.mock()
+    def test_get_survey_list_no_surveys(self, mock_request):
+        mock_request.get(url_get_survey_list, status_code=204)
+
+        response = self.app.get('/backstage-api/v1/survey/surveys')
+
+        self.assertEqual(response.status_code, 200)
```
benjefferies: Seems a little inconsistent to return 204 from one service with no data to the return 200. Seems like for consistency we should probably do one or the other?

ras_backstage/error_handlers.py
andrewmil: This is a bug in flask-restplus. Currently the error handling wont work unless you have both of these here

test/resources/test_survey.py
andrewmil: I do kind of agree but doing it this way does simplify the logic in frontstage.
Just returning an empty array to frontstage is nicer than having to check for 200 and 204.

We don't use 204 in any other services I don't think (for some we return 404) so it might be nice to change that to 200 in the survey service for consistency.

andrewmil: One of the reasons for this backstage service was to move logic away from the frontend, this does that

PR #21, opened 2017-12-13
Update logging to match SDC standards

andrewmil:
This repo was pretty close to the standards already, just made a couple of small changes
[trello](https://trello.com/c/soVzqN6L/4-bring-logging-standards-in-line-with-sdc-logging-standards)

### DIFF
--- a/ras_backstage/controllers/secure_messaging_controller.py
+++ b/ras_backstage/controllers/secure_messaging_controller.py
@@ -18,9 +18,7 @@ def get_messages_list(encoded_jwt, label="", limit=1000):
     response = request_handler('GET', url, headers=headers, params={"label": label, "limit": limit})
 
     if response.status_code != 200:
-        logger.error('Error retrieving the messages list',
-                     label=label,
-                     status=response.status_code)
+        logger.error('Error retrieving the messages list', label=label)
         raise ApiError(url, response.status_code)
 
     logger.debug('Successfully retrieved the messages list', label=label)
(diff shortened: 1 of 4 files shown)


ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -54,10 +49,11 @@ def update_label(encoded_jwt, message_id, label, action):
     response = request_handler('PUT', url, headers=headers, json=data)
 
     if not response or response.status_code != 200:
-        logger.error('Error updating label', status=response.status_code,
-                     message_id=message_id, label=label, action=action)
+        logger.error('Error updating label', message_id=message_id, label=label, action=action)
```
rhiGriff: Should we still log the status code here?

ras_backstage/controllers/survey_controller.py
```diff
@@ -14,14 +14,13 @@
...
     url = '{}{}'.format(app.config['RM_SURVEY_SERVICE'], 'surveys')
-
     response = request_handler('GET', url, auth=app.config['BASIC_AUTH'])
 
     if response.status_code == 204:
         logger.debug('No surveys found in survey service')
         return []
     if response.status_code != 200:
-        logger.error('Error retrieving the survey list', status=response.status_code)
+        logger.error('Error retrieving the survey list')
```
rhiGriff: Same here?

ras_backstage/controllers/secure_messaging_controller.py
andrewmil: It's logged out when we raise ApiError.

We can probably change this so that we don't need to log before raising the ApiError. We do this in frontstage-api already as it as has a slightly different ApiError https://github.com/ONSdigital/ras-frontstage-api/blob/master/frontstage_api/error_handlers.py#L27

It'd probably be good to introduce that in backstage too but not as part of this PR

emilioward:
Is there a requirement to capture more specific status codes like 400, 500 etc?

andrewmil:
@emilioward Not currently

I think we'd only want to catch them specifically when the UI is reacting to them differently, .e.g. for 204 from the survey service we display an empty table. At the moment any kind of error just causes a failure/generic error screen

PR #22, opened 2017-12-15
Us002b view collection exercises

andrewmil:
Added endpoint to return the data required for the single survey view page in the internal UI.

This includes survey data (short/longName, surveyRef and legalBasis(not yet implemented))
This also includes a list of collection exercises information

[trello](https://trello.com/c/TZoUzqkn/24-us002b-view-collection-exercises-13pts)

### DIFF
--- a/ras_backstage/controllers/collection_exercise_controller.py
+++ b/ras_backstage/controllers/collection_exercise_controller.py
@@ -1,0 +1,27 @@
+import json
+import logging
+
+from structlog import wrap_logger
+
+from ras_backstage import app
+from ras_backstage.common.requests_handler import request_handler
+from ras_backstage.exception.exceptions import ApiError
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_collection_exercises_by_survey(survey_id):
+    logger.debug('Retrieving collection exercises', survey_id=survey_id)
+    url = f'{app.config["RM_COLLECTION_EXERCISE_SERVICE"]}collectionexercises/survey/{survey_id}'
+    response = request_handler('GET', url, auth=app.config['BASIC_AUTH'])
+
+    if response.status_code == 204:
+        logger.debug('No collection exercises', survey_id=survey_id)
+        return []
+    if response.status_code != 200:
+        logger.error('Error retrieving collection exercises', survey_id=survey_id)
+        raise ApiError(url, response.status_code)
+
+    logger.debug('Successfully retrieved collection exercises', survey_id=survey_id)
+    return json.loads(response.text)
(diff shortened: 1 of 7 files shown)


ras_backstage/resources/survey/get_survey_by_short_name.py
```diff
@@ -0,0 +1,27 @@
...
+from structlog import wrap_logger
+
+from ras_backstage import survey_api
+from ras_backstage.controllers import collection_exercise_controller, survey_controller
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+@survey_api.route('/shortname/<string:short_name>')
```
benjefferies: It's worth bearing in mind if we need to get this information by any other query i.e. survey id we could revisit this API and make it based on query params

test/resources/test_survey.py
```diff
@@ -53,3 +69,35 @@ def test_get_survey_list_JSONDecodeError(self, mock_request):
         self.assertEqual(response.status_code, 500)
         self.assertIn('"json": "aaaa"'.encode(), response.data)
         self.assertIn('"message": "Expecting value"'.encode(), response.data)
+
+    @requests_mock.mock()
```
benjefferies: Maybe worth adding a test for the empty data scenario

andrewmil:
[US002b: View Collection Exercise(s) (13pts)](https://trello.com/c/TZoUzqkn/24-us002b-view-collection-exercises-13pts)

ras_backstage/controllers/collection_exercise_controller.py
```diff
@@ -0,0 +1,27 @@
...
+from ras_backstage.common.requests_handler import request_handler
+from ras_backstage.exception.exceptions import ApiError
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_collection_exercises_by_survey(survey_id):
+    logger.debug('Retrieving collection exercises', survey_id=survey_id)
+    url = f'{app.config["RM_COLLECTION_EXERCISE_SERVICE"]}collectionexercises/survey/{survey_id}'
```
rhiGriff: Shouldn't we be consistent with these in that we either use .format or this method throughout the whole app?

jcox-dev: Consistency is preferred but as f-strings are not intended to replace str.format but provide an alternative short-hand we can have a mix (especially for different use-cases).

"[This PEP does not propose to remove or deprecate any of the existing string formatting mechanisms](https://www.python.org/dev/peps/pep-0498/#abstract)."

PR #23, opened 2017-12-21
Add endpoint for retrieving party details

andrewmil:
Added endpoint for retrieving combined business and respondent party details at `/party/party-details`

Tests to be added


ras_backstage/controllers/party_controller.py
```diff
@@ -0,0 +1,37 @@
...
+from ras_backstage.common.requests_handler import request_handler
+from ras_backstage.exception.exceptions import ApiError
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_party_by_business_id(party_id):
+    logger.debug('Retrieving business party', party_id=party_id)
+    url = f'{app.config["RAS_PARTY_SERVICE"]}party-api/v1/businesses/id/{party_id}'
```
dcdarrell9: This way is interesting instead of .format(), looks good though

ras_backstage/controllers/party_controller.py
```diff
@@ -0,0 +1,37 @@
...
+        logger.error('Error retrieving business party', party_id=party_id)
+        raise ApiError(url, response.status_code)
+
+    logger.debug('Successfully retrieved business party', party_id=party_id)
+    return json.loads(response.text)
+
+
+def get_party_by_respondent_id(party_id):
+    logger.debug('Retrieving respondent party', party_id=party_id)
+    url = f'{app.config["RAS_PARTY_SERVICE"]}party-api/v1/respondents/id/{party_id}'
```
jcox-dev: Might be worth using `urljoin` on config-prepended strings, depending on how accurate they need to be?

andrewmil: I like the idea, this is something we'd probably want to do across a lot of our repos rather than in this particular PR?

andrewmil: I think we're wanting to merge this in to move release 13 on right now but we can look into using `urljoin` in the next batch of changes we make in this type of repo

jcox-dev: Sure thing 👍

PR #24, opened 2017-12-21
Stop tox from installing dependencies multiple times

jcox-dev:


jcox-dev:
@andrewmil I don't think so and perhaps, really, it's worth removing tox entirely if it isn't used in the CI checks but this PR was to make sure those that do use tox are doing so in the correct environment. 

Cheers for spotting the missing env 👍

PR #26, opened 2018-01-04
Us003 view collection exercise details

andrewmil:
Added an endpoint, `/collection-exercise/<short_name>/<period>` which returns json with the survey and collection-exercise data required for the single collection exercise view in response-operations-ui.

To gather this 3 calls must be made
- Get survey data from survey service
- Use the survey id to get a list of collection exercises from the collection exercise service
- Use the collection exercise id for the right period to retrieve the full collection exercise

To test you can go to `localhost:8001/` while running locally and use the generated swagger ui

[Trello](https://trello.com/c/Gsa6v2ou/37-us003-view-collection-exercise-details-l-5-days)
[Confluence](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/160235584/US003+View+Collection+Exercise+Details+L)


test/resources/test_collection_exercise.py
```diff
@@ -0,0 +1,91 @@
...
+
+import requests_mock
+
+from ras_backstage import app
+
+
+url_get_survey_by_short_name = f'{app.config["RM_SURVEY_SERVICE"]}surveys/shortname/test'
+url_ces = '{}collectionexercises/survey/{}'.format(app.config["RM_COLLECTION_EXERCISE_SERVICE"],
+                                                   "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87")
+url_ce = '{}collectionexercises/{}'.format(app.config["RM_COLLECTION_EXERCISE_SERVICE"],
```
gemmai95: Just out of curiosity why have you used a mix of .format() and f for formatting urls here?

andrewmil: I've done this because of line length, it might not be the best practice.

I prefer f-strings in general but I don't think they look better when you have to do them over multiple lines. So when I've been using .format it's usually because they went over the 100 character line length. 

Consistency might be better though so i could change

ras_backstage/common/filters.py
```diff
@@ -0,0 +1,5 @@
+def filter_collection_exercises(exercises, period):
+    for exercise in exercises:
+        if exercise['name'] == period:
+            return exercise
+    return None
```
jcox-dev: Linter will probably complain about the last line here being unnecessary.

ras_backstage/common/filters.py
```diff
@@ -0,0 +1,5 @@
+def filter_collection_exercises(exercises, period):
```
jcox-dev: `filter_collection_exercises_by_period` to be explicit?

ras_backstage/controllers/collection_exercise_controller.py
```diff
@@ -11,6 +11,22 @@
 logger = wrap_logger(logging.getLogger(__name__))
 
 
+def get_collection_exercise_by_id(collection_exercise_id):
+    logger.debug('Retrieving collection exercise', collection_exercise_id=collection_exercise_id)
+    url = '{}collectionexercises/{}'.format(app.config["RM_COLLECTION_EXERCISE_SERVICE"],
```
jcox-dev: f-string or [urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) would probably be clearer here.

ras_backstage/common/filters.py
jcox-dev: also seems to be a get rather than a filter as it returns on the first found

ras_backstage/common/filters.py
andrewmil: Does a function `return None` as default if you don't hit a return statement? I'll remove that if so

jcox-dev: Yeah, without an explicit return statement all Python methods and function return None by default.

andrewmil:
[US003: View Collection Exercise Details (L = 5 days)](https://trello.com/c/Gsa6v2ou/37-us003-view-collection-exercise-details-l-5-days)

PR #29, opened 2018-01-09
US009 Upload sample to rm-sample-service

jcox-dev:
Adds a `/backstage-api/v1/sample/<short_name>/<period>` POST endpoint for uploading sample files to the [sample service](https://github.com/ONSdigital/rm-sample-service) and creating a link to the collection exercise when successful for retrieval later.


jcox-dev:
This pattern of accepting uploads on the frontend, which are then passed on to backstage before the RM services is consistent with [what we do already for collection instruments](https://github.com/ONSdigital/ras-frontstage/blob/master/frontstage/views/surveys.py#L124).

ras_backstage/controllers/sample_controller.py
```diff
@@ -0,0 +1,40 @@
...
+
+from ras_backstage import app
+from ras_backstage.common.requests_handler import request_handler
+from ras_backstage.exception.exceptions import ApiError
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def post_sample_file_for_collection_exercise(collection_exercise_id, filename, sample_file_data, survey_type='bres'):
```
andrewmil: What's the reason for having `bres` as a default?

ras_backstage/resources/sample/post_sample_file.py
```diff
@@ -0,0 +1,38 @@
...
+parser.add_argument('collection_exercise_id', location='args', required=True)
+
+
+@sample_api.route('/upload')
+class PostSampleFile(Resource):
+
+    @staticmethod
+    @sample_api.expect(parser)
+    def post():
+        collection_exercise_id = request.args['collection_exercise_id']
```
andrewmil: The parser doesn't actually enforce the `required=True` for query string arguments so if you try and post without `?collection_exercise_id=asd....` in the url it will fail on this line.

You do still get a 400 response as is right now, the error message just wont be that helpful

Would this endpoint work better with endpoint parameters? i.e. `/upload/<collection_exercise_id>`

ras_backstage/controllers/sample_controller.py
jcox-dev: I think I mistakenly thought that was what the API meant by "survey type". Is it actually "B" / "BI"?

ras_backstage/resources/sample/post_sample_file.py
jcox-dev: If it doesn't enforce them then it's probably worth using a parameter instead. 👍

ras_backstage/controllers/sample_controller.py
andrewmil: I'm not sure :/
Just had a quick look in the survey service and I think this is where we'd specify it was a business survey but I can't find anything that makes it that clear

andrewmil: Had another look at the sample service. Looks like every sample thats in it currently has type B.
Also the endpoint is at `samples/{survey_type}/fileupload` the Api.md for sample service doesn't make that very clear

ras_backstage/__init__.py
```diff
@@ -22,33 +22,36 @@
 
 collection_exercise_api = Namespace('collection-exercise',
                                     path='/backstage-api/v1/collection-exercise')
+collection_instrument_api = Namespace('survey', path='/backstage-api/v1/collection-instrument')
```
andrewmil: This isn't originally from this PR but this Namespace should be called `'collection-instrument'` rather than `'survey'`

ras_backstage/resources/sample/sample.py
```diff
@@ -0,0 +1,31 @@
...
+
+from ras_backstage import sample_api
+from ras_backstage.common.filters import get_collection_exercise_by_period
+from ras_backstage.controllers import collection_exercise_controller, sample_controller, survey_controller
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+@sample_api.route('/<short_name>/<period>')
+class Sample(Resource):
```
andrewmil: There's a small amount of extra code that can be added here to make the swagger api accept files and (I think) validate that a file is required for this request. There's an example of it in [frontstage-api](https://github.com/ONSdigital/ras-frontstage-api/blob/master/frontstage_api/resources/surveys/upload_collection_instrument.py#L16)
This could also be done for the upload collection instrument

config.py
```diff
@@ -63,6 +63,13 @@ class Config(object):
                                              RM_SURVEY_SERVICE_HOST,
                                              RM_SURVEY_SERVICE_PORT)
 
+    RM_SAMPLE_SERVICE_HOST = os.getenv('RM_SAMPLE_SERVICE_HOST', 'localhost')
+    RM_SAMPLE_SERVICE_PORT = os.getenv('RM_SAMPLE_SERVICE_PORT', 8125)
```
gemmai95: Have these been added to the readme?

jcox-dev: The readme does not currently mention config variables.

andrewmil: I had a chat with @benjefferies about this a while back and the idea was that rather than writing out all the environment variables in the readme we'd just have a link to the config file

gemmai95: Has this been made aware to Dave or Mark?

davidmort: i definitely think that we should be listing the the environment variables in the readme (at least the mandatory ones), the readme should give all info required to setup/startup the service

config.py
```diff
@@ -63,6 +63,13 @@ class Config(object):
...
+    RM_SAMPLE_SERVICE_PROTOCOL = os.getenv('RM_SAMPLE_SERVICE_PROTOCOL', 'http')
+    RM_SAMPLE_SERVICE = '{}://{}:{}/'.format(RM_SAMPLE_SERVICE_HOST,
+                                             RM_SAMPLE_SERVICE_PORT,
+                                             RM_SAMPLE_SERVICE_PROTOCOL)
```
tej99: RM_SAMPLE_SERVICE_PROTOCOL is in the wrong place, need to be the first one.

jcox-dev: Hah, good spot.

jcox-dev:
@gemmai95 is there a template to follow for listing environment variables we can use here?

config.py
benjefferies: @davidmort it looks like the environmental variables documented in the readme is just a duplication of what is in the readme now. Also moves away from one source of the truth.

Would a preferable approach would be that the app works out the box and therefore doesn't documentation to get up and running?

ras_backstage/resources/collection_instrument/collection_instrument.py
```diff
@@ -1,21 +1,26 @@
...
 
 from ras_backstage import collection_instrument_api
 from ras_backstage.common.filters import get_collection_exercise_by_period
 from ras_backstage.controllers import survey_controller, collection_exercise_controller, \
     collection_instrument_controller
 
+
 logger = wrap_logger(logging.getLogger(__name__))
 
+parser = reqparse.RequestParser()
```
benjefferies: 👍

ras_backstage/resources/sample/sample.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @staticmethod
+    @sample_api.expect(parser)
+    def post(short_name, period):
+        logger.info('Uploading sample', shortname=short_name, period=period)
+        survey = survey_controller.get_survey_by_shortname(short_name)
+        # TODO: get survey type from service(?)
```
benjefferies: Is this todo still valid?

ras_backstage/controllers/sample_controller.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def upload_sample(survey_id, collection_exercise_id, sample_file, survey_type='B'):
+    logger.debug('Uploading sample file',
+                 collection_exercise_id=collection_exercise_id,
+                 survey_id=survey_id,
+                 survey_type=survey_type)
+    url = f'{app.config["RM_SAMPLE_SERVICE"]}samples/{survey_type}/fileupload'
+
+    response = request_handler(url=url, method='POST', auth=app.config['BASIC_AUTH'], files={'file': sample_file})
```
benjefferies: I found it wasn't actually posting the file so had to do something like https://github.com/ONSdigital/ras-backstage/pull/36

ras_backstage/resources/sample/sample.py
jcox-dev: No, I don't think it is as we're currently only handling business (B) surveys.

ras_backstage/controllers/sample_controller.py
jcox-dev: Fix #36 not needed here.

ras_backstage/controllers/sample_controller.py
```diff
@@ -0,0 +1,37 @@
...
+
+from ras_backstage import app
+from ras_backstage.common.requests_handler import request_handler
+from ras_backstage.exception.exceptions import ApiError
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def upload_sample(survey_id, collection_exercise_id, sample_file, survey_type='B'):
```
jcox-dev: Issue found in sample service, which incorrectly expects a survey type of ["bres" / "social" / "census"](https://github.com/ONSdigital/rm-sample-service/blob/da3b122b3696e64c28af872c38f720af2f6337df/src/main/java/uk/gov/ons/ctp/response/sample/service/impl/SampleServiceImpl.java#L234) and not "B". The service also does not raise an error upon [receipt of an unexpected type](https://github.com/ONSdigital/rm-sample-service/blob/da3b122b3696e64c28af872c38f720af2f6337df/src/main/java/uk/gov/ons/ctp/response/sample/service/impl/SampleServiceImpl.java#L245) and still returns an empty 201.

ras_backstage/resources/sample/sample.py
```diff
@@ -5,32 +5,33 @@
...
+        collection_exercise_id = request.args.get('collection_exercise_id')
+        if not collection_exercise_id:
+            return make_response(jsonify({'message': 'Missing collection_exercise_id'}), 400)
 
-        exercises = collection_exercise_controller.get_collection_exercises_by_survey(survey['id'])
-        # Find the collection exercise for the given period
-        exercise = get_collection_exercise_by_period(exercises, period)
+        logger.info('Uploading sample', collection_exercise_id=collection_exercise_id)
+
+        exercise = collection_exercise_controller.get_collection_exercise_by_id(collection_exercise_id)
```
tej99: is this check needed? does sample service not have the validation in?

jcox-dev: @tej99 
The check may not be needed but the sample service has no knowledge of the collection exercise that the sample is being uploaded against (until, presumably, the file is processed). Perhaps it's worth guarding against uploading a sample that is for a currently unknown CE?

If we do hand over validation to the sample service, the sample upload resource probably doesn't need any parameters – survey short name, collection exercise or otherwise.

config.py
andrewmil: @benjefferies I think as it is the application will work out of the box but will try to connect to other services on localhost?
There's a lot of cases currently for developers where they'd need to change that though

jcox-dev:
@andrewmil we're waiting on Ben's fix in the sample service (BRES -> B) before this can work as expected.

jcox-dev:
@gemmai95 environment variables are now in the [README](https://github.com/ONSdigital/ras-backstage/blob/upload-to-sample-service/README.md).

PR #31, opened 2018-01-10
Adding upload from backstage to collection instrument

benjefferies:


ras_backstage/controllers/collection_instrument_controller.py
```diff
@@ -0,0 +1,34 @@
...
+    logger.debug('Uploading collection instrument', collection_exercise_id=collection_exercise_id)
+    url = f'{app.config["RAS_COLLECTION_INSTRUMENT_SERVICE"]}upload/{collection_exercise_id}'
+
+    classifiers = {}
+    if survey_id:
+        classifiers['SURVEY_ID'] = survey_id
+    if collection_exercise_id:
+        classifiers['COLLECTION_EXERCISE'] = collection_exercise_id
+    try:
+        response = requests.post(url, auth=app.config['BASIC_AUTH'], files={'file': file})
```
andrewmil: I saw this is WIP but it's probably worth mentioning that we have a `request_handler` function in `ras_backstage/common/requests_handler` which can handle the `RequestException`

ras_backstage/controllers/collection_instrument_controller.py
```diff
@@ -0,0 +1,27 @@
...
+    if survey_id:
+        classifiers['SURVEY_ID'] = survey_id
+    if collection_exercise_id:
+        classifiers['COLLECTION_EXERCISE'] = collection_exercise_id
+
+    request_handler(url=url, method='POST', auth=app.config['BASIC_AUTH'], files={'file': file},
+                    params={'classifiers': json.dumps(classifiers)})
+
+    logger.debug('Successfully uploaded collection instrument',
+                 collection_exercise_id=collection_exercise_id)
```
andrewmil: Missing newline

ras_backstage/controllers/collection_instrument_controller.py
```diff
@@ -0,0 +1,27 @@
...
+from structlog import wrap_logger
+
+from ras_backstage import app
+from ras_backstage.common.requests_handler import request_handler
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def upload_collection_instrument(survey_id, collection_exercise_id, file):
+    logger.debug('Uploading collection instrument', collection_exercise_id=collection_exercise_id)
```
andrewmil: Would it be worth including the `survey_id` in the log message too since we have it here? 
The more info we can be logging out the better probably

ras_backstage/resources/collection_instrument/collection_instrument.py
```diff
@@ -0,0 +1,29 @@
...
+        logger.info('Uploading collection instrument', shortname=short_name, period=period)
+        survey = survey_controller.get_survey_by_shortname(short_name)
+        exercises = collection_exercise_controller.get_collection_exercises_by_survey(survey['id'])
+        # Find the collection exercise for the given period
+        exercise = get_collection_exercise_by_period(exercises, period)
+        if not exercise:
+            return make_response(jsonify({"message": "Collection exercise not found"}), 404)
+
+        collection_instrument_controller.upload_collection_instrument(survey['id'], exercise['id'],
+                                                                      request.files['file'])
```
andrewmil: For consistency would be good to have an info level successful log message at the end of this function

ras_backstage/resources/collection_instrument/collection_instrument.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @staticmethod
+    def post(short_name, period):
+        logger.info('Uploading collection instrument', shortname=short_name, period=period)
+        survey = survey_controller.get_survey_by_shortname(short_name)
+        exercises = collection_exercise_controller.get_collection_exercises_by_survey(survey['id'])
+        # Find the collection exercise for the given period
+        exercise = get_collection_exercise_by_period(exercises, period)
+        if not exercise:
+            return make_response(jsonify({"message": "Collection exercise not found"}), 404)
```
andrewmil: We should have a test for this error case

benjefferies: good spot

ras_backstage/resources/collection_instrument/collection_instrument.py
benjefferies: done

ras_backstage/controllers/collection_instrument_controller.py
```diff
@@ -0,0 +1,27 @@
...
+from ras_backstage import app
+from ras_backstage.common.requests_handler import request_handler
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def upload_collection_instrument(survey_id, collection_exercise_id, file):
+    logger.debug('Uploading collection instrument', collection_exercise_id=collection_exercise_id, survey_id=survey_id)
+    url = f'{app.config["RAS_COLLECTION_INSTRUMENT_SERVICE"]}' \
+          f'collection-instrument-api/1.0.2/upload/{collection_exercise_id}'
```
jcox-dev: Is the CE API version likely to change? If so, it might be worth adding it as a config / env variable.

benjefferies: Unsure but I'd guess if it was to change it would be because of a breaking change so might need code changes anyway and at that point is it worth being an environmental variable?

PR #34, opened 2018-01-12
Fix/sort survey list by survey ref

andrewmil:
Sort survey list by surveyRef.

From missed acceptance criteria in [confluence](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/166920222/US001b+View+Survey+Details)

### DIFF
--- a/ras_backstage/common/mappers.py
+++ b/ras_backstage/common/mappers.py
@@ -2,5 +2,4 @@ import re
 
 
 def format_short_name(short_name):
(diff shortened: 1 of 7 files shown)


andrewmil:
@benjefferies Added that test. We should probably assert this in the integration tests as well, I've mentioned it to @rhiGriff

rhiGriff:
@andrewmil it will be added to the integration tests as discussed

PR #36, opened 2018-01-16
Fix file upload bug

benjefferies:
WIP: Investigating how we can test that the files are uploaded correctly

The file was not being uploaded correctly and the filename was not being
set.

https://trello.com/c/jHAVOeuA/32-us006-load-seft-collection-instruments-25pts-l-5-days


jcox-dev:
@benjefferies likely to affect upload of sample as well?

PR #37, opened 2018-01-22
US010 Add sample summary to collection exercise

jcox-dev:
Adds controllers for including a collection exercise's sample in the GET response.


jcox-dev:
Codacy issue looks like a bug their side.

jcox-dev:
[US010: View sample in context of a CE](https://trello.com/c/9zdn4R3Q/26-us010-view-sample-in-context-of-a-ce)

ras_backstage/controllers/collection_exercise_controller.py
```diff
@@ -41,3 +41,32 @@ def get_collection_exercises_by_survey(survey_id):
...
+    elif response.status_code == 204:
+        logger.error('No samples linked to collection exercise', collection_exercise_id=collection_exercise_id)
+        raise ApiError(url, response.status_code)
+    elif response.status_code != 200:
+        logger.error('Error retrieving sample summaries linked to collection exercise',
+                     collection_exercise_id=collection_exercise_id)
+        raise ApiError(url, response.status_code)
+
+    try:
+        sample_summary_id = response.json()[0]
```
andrewmil: Do we need the try-except here. If there's no samples linked shouldn't that be caught with the 204 status code?

ras_backstage/controllers/sample_controller.py
```diff
@@ -0,0 +1,28 @@
...
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_sample_summary(sample_summary_id):
+    logger.debug('Retrieving sample summary', sample_summary_id=sample_summary_id)
+    url = f'{app.config["RM_SAMPLE_SERVICE"]}samples/samplesummary/{sample_summary_id}'
+
+    response = request_handler(url=url, method='GET', auth=app.config['BASIC_AUTH'])
+
+    # Sample service *should* return something other than 201 when upload / ingest fails
```
andrewmil: This comment is copy pasted from somewhere else, can probably be removed

ras_backstage/controllers/collection_exercise_controller.py
jcox-dev: @andrewmil it *should* be but I'd rather catch a sudden change to the API and return a 404 than cause an unexpected exception to bubble up. Unless that's preferred?

jcox-dev:
@andrewmil agreed, the additional endpoint is probably unncessary. I'll amend to append the sample summary to the single collection exercise.

ras_backstage/controllers/collection_exercise_controller.py
```diff
@@ -41,3 +41,26 @@ def get_collection_exercises_by_survey(survey_id):
...
     return json.loads(response.text)
+
+
+def get_linked_sample_summary_id(collection_exercise_id):
+    logger.debug('Retrieving sample linked to collection exercise', collection_exercise_id=collection_exercise_id)
+    url = f'{app.config["RM_COLLECTION_EXERCISE_SERVICE"]}collectionexercises/link/{collection_exercise_id}'
+    response = request_handler('GET', url, auth=app.config['BASIC_AUTH'])
+
+    if response.status_code == 204:
+        logger.error('No samples linked to collection exercise', collection_exercise_id=collection_exercise_id)
```
tej99: This should just return None instead of raising an api error because if there is no linked samples we still want the collection exercise details page to load just with an upload sample option.

ras_backstage/resources/collection_exercise/single_collection_exercise.py
```diff
@@ -33,10 +33,14 @@ def get(short_name, period):
 
         collection_instruments = get_collection_instruments_by_classifier(survey['id'], exercise['id'])
 
+        summary_id = collection_exercise_controller.get_linked_sample_summary_id(exercise['id'])
+        sample_summary = sample_controller.get_sample_summary(summary_id)
```
tej99: sample_controller.get_sample_summary(summary_id) if summary_id else None 
or any other indicator to the ui that there is no sample_summary

PR #39, opened 2018-01-23
Usxxx internal user sign in

insacuri:
Putting in a sign-in endpoint that will be used by response-operations-ui.  At the moment it's basically a mock endpoint so that we have something to point the login page at.  Once UAA is implemented, this will change drastically.

We're assuming that uaa will give an oauth2 token at the end of it all so we'll pretend to pass that back, but again, this will probably be very different once it's implemented


ras_backstage/resources/sign_in_uaa/sign_in_uaa.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@sign_in_uaa_api.route('/', methods=['POST'])
+class SignInUaa(Resource):
+
+    @staticmethod
+    @sign_in_uaa_api.expect(sign_in_details, validate=True)
+    def post():
+        logger.info('Retrieving sign-in details')
+        # force=true means there post doesn't HAVE to have application/json
```
JamesGardiner: typo in docstring - there should be the I think

ras_backstage/resources/sign_in_uaa/sign_in_uaa.py
```diff
@@ -0,0 +1,39 @@
...
+        # in its content type.  This feels wrong to me but is also in the django signin code...
+        message_json = request.get_json(force=True)
+        username = message_json['username']
```
JamesGardiner: use `.get` to avoid KeyError being raised

ras_backstage/resources/sign_in_uaa/sign_in_uaa.py
insacuri: Fixed

ras_backstage/resources/sign_in_uaa/sign_in_uaa.py
insacuri: Fixed

ras_backstage/resources/sign_in_uaa/sign_in_uaa.py
```diff
@@ -0,0 +1,39 @@
...
+        'password': fields.String(required=True, description='password')
+})
+
+
+@sign_in_uaa_api.route('/', methods=['POST'])
+class SignInUaa(Resource):
+
+    @staticmethod
+    @sign_in_uaa_api.expect(sign_in_details, validate=True)
+    def post():
```
andrewmil: I understand that this isn't functionally doing exactly what we want right now but it'd probably be a good idea to add a simple test for it so it doesn't get forgotten

ras_backstage/__init__.py
```diff
@@ -27,6 +27,7 @@
 sample_api = Namespace('sample', path='/backstage-api/v1/sample')
 secure_messaging_api = Namespace('secure-messaging', path='/backstage-api/v1/secure-message')
 sign_in_api = Namespace('sign-in', path='/backstage-api/v1/sign-in')
+sign_in_uaa_api = Namespace('sign-in-uaa', path='/backstage-api/v1/sign-in-uaa')
```
warrenbailey: Do we need to support both?

I'd rather get rid of the old one, as we'll replace it eventually anyway.

If we do then put it as v2?
i.e.
/backstage-api/v2/sign-in-uaa

ras_backstage/resources/sign_in_uaa/sign_in_uaa.py
```diff
@@ -0,0 +1,39 @@
...
+        logger.info('Retrieving sign-in details')
+        # force=true means the post doesn't HAVE to have application/json
+        # in its content type.  This feels wrong to me but is also in the django signin code...
```
warrenbailey: Pleased remove this comment, either don't use force=True and handle the None reponse yourself or leave as is and use the on_json_loading_failed function instead. The default implementation will raise a bad request if the request body can't be parsed to json

http://flask.pocoo.org/docs/0.12/api/#flask.Request.get_json

ras_backstage/resources/sign_in_uaa/sign_in_uaa.py
JamesGardiner: @andrewmil test added

ras_backstage/__init__.py
andrewmil: We need to support both right now.
It's being used by the angular application which is still in use for BRES

ras_backstage/resources/sign_in_uaa/sign_in_uaa.py
andrewmil: There should be no need to be using `force=True` here anyway. We're in control of the only place which will make a request to here so we know what form it's coming in as

warrenbailey: Agreed

JamesGardiner: I agree with @andrewmil - I'll take it out.

PR #40, opened 2018-01-29
Remove unnecessary classifiers survey_id and collection_exercise_id

LJBabbage:
Survey_id and Collection_exercise_id shouldn't be added to the classifiers JSON. They are already formed tables with data and relationships in collection instrument. By adding them into the classifiers table, we are not only causing duplication but also causing issues when a search is made with them. They need to be removed


JamesGardiner:
@tej99 AND @LJBabbage - this pr is breaking tests. Can you revert it or add a fix?

PR #44, opened 2018-02-02
usxxx internal user sign in with uaa

JamesGardiner:
This PR makes the use of UAA configurable via the app's environment. If USE_UAA is set in the app's environment then the app will attempt to use UAA to authenticate. This option is only in place to allow local dev to continue whilst a local UAA mock/docker solution is in place. In future iterations UAA will be the only authentication mechanism.

### DIFF
--- a/config.py
+++ b/config.py
@@ -17,8 +17,8 @@ class Config(object):
     JWT_SECRET = os.getenv('JWT_SECRET')
 
     # TODO: remove once UAA is implemented and use user supplied username/password
-    USERNAME = os.getenv('RAS_BACKSTAGE_USERNAME', 'user')
-    PASSWORD = os.getenv('RAS_BACKSTAGE_PASSWORD', 'pass')
+    USERNAME = os.getenv('RAS_BACKSTAGE_USERNAME')
+    PASSWORD = os.getenv('RAS_BACKSTAGE_PASSWORD')
 
     RAS_OAUTH_SERVICE_HOST = os.getenv('RAS_OAUTH_SERVICE_HOST', 'localhost')
     RAS_OAUTH_SERVICE_PORT = os.getenv('RAS_OAUTH_SERVICE_PORT', 8040)
@@ -91,6 +91,8 @@ class DevelopmentConfig(Config):
     DJANGO_CLIENT_SECRET = os.getenv('DJANGO_CLIENT_SECRET', 'password')
     DJANGO_BASIC_AUTH = (DJANGO_CLIENT_ID, DJANGO_CLIENT_SECRET)
     JWT_SECRET = os.getenv('JWT_SECRET', 'testsecret')
+    USERNAME = os.getenv('RAS_BACKSTAGE_USERNAME', 'user')
+    PASSWORD = os.getenv('RAS_BACKSTAGE_PASSWORD', 'pass')
 
 
 class TestingConfig(DevelopmentConfig):
--- a/config.py
+++ b/config.py
@@ -79,7 +79,7 @@ class Config(object):
     UAA_SERVICE_URL = os.getenv('UAA_SERVICE_URL', 'localhost')
     UAA_CLIENT_ID = os.getenv('UAA_CLIENT_ID', 'ras_backstage_client_id')
     UAA_CLIENT_SECRET = os.getenv('UAA_CLIENT_SECRET', 'password')
-    USE_UAA = os.getenv('USE_UAA', default=True)
+    USE_UAA = os.getenv('USE_UAA', True)
 
 
 class DevelopmentConfig(Config):
--- a/config.py
+++ b/config.py
@@ -79,6 +79,7 @@ class Config(object):
     UAA_SERVICE_URL = os.getenv('UAA_SERVICE_URL', 'localhost')
     UAA_CLIENT_ID = os.getenv('UAA_CLIENT_ID', 'ras_backstage_client_id')
     UAA_CLIENT_SECRET = os.getenv('UAA_CLIENT_SECRET', 'password')
+    USE_UAA = os.getenv('USE_UAA', default=True)
 
 
 class DevelopmentConfig(Config):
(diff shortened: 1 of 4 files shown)


ras_backstage/controllers/uaa_controller.py
```diff
@@ -0,0 +1,40 @@
...
+from ras_backstage.common.requests_handler import request_handler
+from ras_backstage.exception.exceptions import ApiError
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def sign_in(username, password):
+    logger.debug('Retrieving OAuth2 token for sign-in')
+    url = '{}{}'.format(app.config['UAA_SERVICE_URL'], '/oauth/token')
```
dcdarrell9: Shall we use an f'string here for consistency?

config.py
```diff
@@ -76,6 +76,11 @@ class Config(object):
                                              RM_SAMPLE_SERVICE_HOST,
                                              RM_SAMPLE_SERVICE_PORT)
 
+    UAA_SERVICE_URL = os.getenv('UAA_SERVICE_URL', 'localhost')
+    UAA_CLIENT_ID = os.getenv('UAA_CLIENT_ID', 'ras_backstage_client_id')
+    UAA_CLIENT_SECRET = os.getenv('UAA_CLIENT_SECRET', 'password')
+    USE_UAA = os.getenv('USE_UAA', True)
```
jcox-dev: Will this ever be False?

ras_backstage/resources/sign_in/sign_in.py
```diff
@@ -53,12 +53,28 @@ def post():
...
 
-        # Obviously horrible, stopgap until uaa is implemented
-        if username == current_app.config['USERNAME'] and password == current_app.config['PASSWORD']:
-            # We're assuming that uaa will return an Oauth2 token though it's almost certain that
-            # this will change once we know exactly what is being returned.
-            logger.info("Authentication successful", user=username)
-            return make_response(jsonify({"token": "1234abc"}), 201)
+        if not current_app.config.get('USE_UAA'):
+            logger.info('Retrieving sign-in details')
+            if username == current_app.config['USERNAME'] and password == current_app.config['PASSWORD']:
```
jcox-dev: If no username and password is provided and we're not using the dev config would this evaluate to True (`None == None and None == None`) if the config isn't set correctly?

dcdarrell9: There's validation to make sure a user cannot send no username or password, but might be worth having a test for something like this 👍

insacuri: Form validation kicks in before you get to this point which stops you from submitting nothing

config.py
JamesGardiner: Ah good catch. Happy to wrap it in a `bool()`?

ras_backstage/resources/sign_in/sign_in.py
JamesGardiner: @jcox-dev yes

ras_backstage/resources/sign_in/sign_in.py
```diff
@@ -53,12 +53,28 @@ def post():
...
-            logger.info("Authentication failed", user=username)
-            return make_response(jsonify({"error": "Username and/or password incorrect"}), 401)
+            logger.info('Retrieving sign-in details using UAA')
+            message_json = request.get_json()
+            username = message_json.get('username')
+            password = message_json.get('password')
+
+            oauth2_token = uaa_controller.sign_in(username, password)
+            token_expiry = datetime.now() + timedelta(seconds=int(oauth2_token['expires_in']))
+            oauth2_token['expires_at'] = token_expiry.timestamp()
```
andrewmil: Can we set an expiry time in the uaa service or do we have to set it here?

ras_backstage/resources/sign_in/sign_in.py
```diff
@@ -53,12 +53,28 @@ def post():
...
+            if username == current_app.config['USERNAME'] and password == current_app.config['PASSWORD']:
+                logger.info("Authentication successful", user=username)
+                return make_response(jsonify({"token": "1234abc"}), 201)
+            else:
+                logger.info("Authentication failed", user=username)
+                return make_response(jsonify({"error": "Username and/or password incorrect"}), 401)
         else:
-            logger.info("Authentication failed", user=username)
-            return make_response(jsonify({"error": "Username and/or password incorrect"}), 401)
+            logger.info('Retrieving sign-in details using UAA')
```
andrewmil: We should probably add tests for signing in with UAA

test/resources/test_collection_instrument.py
```diff
@@ -39,14 +37,9 @@ def setUp(self):
...
     def test_upload_collection_instrument(self, mock_request):
-        classifiers = {
-            "SURVEY_ID": "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87",
-            "COLLECTION_EXERCISE": "e33daf0e-6a27-40cd-98dc-c6231f50e84a"
-        }
-        params = urlencode({'classifiers': json.dumps(classifiers)})
         mock_request.get(url_get_survey_by_short_name, json=self.survey)
         mock_request.get(url_ces, json=self.collection_exercises)
-        mock_request.post(f'{url_upload_collection_instrument}?{params}', complete_qs=True,
+        mock_request.post(f'{url_upload_collection_instrument}', complete_qs=True,
```
andrewmil: If we're removing `?{params}` here then we no longer need `complete_qs=True`

ras_backstage/resources/sign_in/sign_in.py
warrenbailey: None of this is actually needed, we're going to use the token directly from the UAA

JamesGardiner:
Closing pull request for now - UAA changes will mean significant changes to the code in this branch. Will reopen once uaa implemented and we have a stable interface.

PR #48, opened 2018-02-08
Us043 find an ru

tej99:
As an: 
internal User
I need: 
to be able to 'find' a Reporting Unit (RU Ref for Business Surveys)
So that: 
I can easily locate the RU to deal with any issues

Dependencies:
https://github.com/ONSdigital/ras-party/pull/99


tej99:
[US043: INT Find an RU [Large]](https://trello.com/c/WyRWFKXq/38-us043-int-find-an-ru-large)

test/resources/test_collection_instrument.py
```diff
@@ -39,15 +37,9 @@ def setUp(self):
...
-        classifiers = {
-            "SURVEY_ID": "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87",
-            "COLLECTION_EXERCISE": "e33daf0e-6a27-40cd-98dc-c6231f50e84a"
-        }
-        params = urlencode({'classifiers': json.dumps(classifiers)})
         mock_request.get(url_get_survey_by_short_name, json=self.survey)
         mock_request.get(url_ces, json=self.collection_exercises)
-        mock_request.post(f'{url_upload_collection_instrument}?{params}', complete_qs=True,
-                          additional_matcher=self.file_matcher)
+        mock_request.post(f'{url_upload_collection_instrument}', additional_matcher=self.file_matcher)
```
andrewmil: :+1:

run_tests.py
```diff
@@ -1,7 +1,9 @@
 import os
+import sys
 
 import pytest
 
 if __name__ == "__main__":
     os.environ['APP_SETTINGS'] = 'TestingConfig'
-    pytest.main(['--cov-report', 'term-missing', '--cov', 'ras_backstage'])
+    exitcode = pytest.main(['--cov-report', 'term-missing', '--cov', 'ras_backstage'])
+    sys.exit(exitcode)
```
andrewmil: :+1:

ras_backstage/resources/reporting_units/search_reporting_units.py
```diff
@@ -0,0 +1,30 @@
...
+    @staticmethod
+    @reporting_unit_api.expect(parser)
+    def get():
+        query = request.args.get('query')
+
+        logger.info('Retrieving reporting units by search query', query=query)
+
+        businesses = party_controller.get_businesses_by_search(query)
+
+        logger.info('Successfully retrieved reporting units by search query')
```
necrophonic: Without context this log line won't usefully match back up with the previous one

ras_backstage/controllers/party_controller.py
```diff
@@ -35,3 +35,16 @@ def get_party_by_respondent_id(party_id):
...
     logger.debug('Successfully retrieved respondent party', party_id=party_id)
     return json.loads(response.text)
+
+
+def get_businesses_by_search(query):
+    logger.debug('Retrieving businesses by search query', query=query)
+    url = f'{app.config["RAS_PARTY_SERVICE"]}party-api/v1/businesses/search'
+    response = request_handler('GET', url, auth=app.config['BASIC_AUTH'], params={"query": query})
+
+    if response.status_code != 200:
```
JamesGardiner: I mentioned this in some other reviews today. I really think we should be using raise_for_status and handling any error codes in the > 300 range.

run_tests.py
JamesGardiner: is there any value in running the tests via a Python module, rather than using the recommended test runner that comes with PyTest?

tej99: I think this way was introduced in secure messaging because the tests ran both unit tests and the behave tests. Then it creeped it's way into other repos as we didn't use make files before, could just run it with the pytest command. I've just fixed the script to fail when tests fail here as it didn't do so previously. To remove the script is probably out of this tickets scope as there are at least three other repos like this.

ras_backstage/controllers/party_controller.py
tej99: In this particular case we can use raise_for_status but there are certain scenarios with error codes in the > 300 range that should not actually raise any errors because the backstage is used as an aggregator. So this is just the standard way we handle responses throughout this application.

PR #49, opened 2018-02-08
Us044 display ru details

andrewmil:
Added new endpoint at `/reporting-units/<ru_ref>`
It will return the data required for the individual reporting unit page
Currently it only returns the party for the given ru_ref

[Trello](https://trello.com/c/zHV82xsH/77-us044-int-view-ru-details-med)
[Confluence](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/248643696/US044+View+RU+Details)


andrewmil:
[US044: INT View RU Details [Med]](https://trello.com/c/zHV82xsH/77-us044-int-view-ru-details-med)

ras_backstage/controllers/party_controller.py
```diff
@@ -21,7 +20,7 @@ def get_party_by_business_id(party_id):
         raise ApiError(url, response.status_code)
 
     logger.debug('Successfully retrieved business party', party_id=party_id)
-    return json.loads(response.text)
+    return response.json()
```
JamesGardiner: You may want to handle ValueErrors gracefully here.

ras_backstage/controllers/party_controller.py
```diff
@@ -34,4 +33,17 @@ def get_party_by_respondent_id(party_id):
         raise ApiError(url, response.status_code)
 
     logger.debug('Successfully retrieved respondent party', party_id=party_id)
-    return json.loads(response.text)
+    return response.json()
```
JamesGardiner: And here

ras_backstage/controllers/party_controller.py
```diff
@@ -34,4 +33,17 @@ def get_party_by_respondent_id(party_id):
...
+
+
+def get_party_by_ru_ref(ru_ref):
+    logger.debug('Retrieving reporting unit', ru_ref=ru_ref)
+    url = f'{app.config["RAS_PARTY_SERVICE"]}party-api/v1/parties/type/B/ref/{ru_ref}'
+    response = request_handler('GET', url, auth=app.config['BASIC_AUTH'])
+
+    if response.status_code != 200:
```
JamesGardiner: We should be using response.raise_for_status and allowing specific error handlers to handle cases where responses aren't in the 200 range.

JamesGardiner: I'm really not a fan of the generic ApiError, it doesn't give us any scope for graceful degradation of service

ras_backstage/controllers/party_controller.py
```diff
@@ -34,4 +33,17 @@ def get_party_by_respondent_id(party_id):
...
+        logger.error('Error retrieving reporting unit', ru_ref=ru_ref)
+        raise ApiError(url, response.status_code)
+
+    logger.debug('Successfully retrieved reporting unit', ru_ref=ru_ref)
+    return response.json()
```
JamesGardiner: Value errors.

ras_backstage/controllers/party_controller.py
andrewmil: My initial reason to add ApiError in was to make it easier to debug which external call has failed in backstage

For quite a few of the backstage endpoints it makes several external calls and the ApiError handler makes it very clear which service it has failed to call

I'm open to removing it though, it would probably need it's own PR though to do that across the repo as it is pretty widespread

PR #53, opened 2018-02-15
add respondents for ru per survey

tej99:
As an: internal user
I need: to be able to view respondent data for a Reporting Unit for a specific survey
So that: I understand the relationships with respondents for the Reporting Unit across surveys

### DIFF
--- a/ras_backstage/resources/reporting_units/get_reporting_unit.py
+++ b/ras_backstage/resources/reporting_units/get_reporting_unit.py
@@ -26,6 +26,8 @@ class GetReportingUnit(Resource):
 
         surveys = get_surveys_for_reporting_unit(reporting_unit)
 
+        respondents_for_survey = get_respondents_for_reporting_unit(reporting_unit)
+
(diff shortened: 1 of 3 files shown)


tej99:
[US046: INT View Enrolled Respondents for a Survey for an RU[Med]](https://trello.com/c/DkC0zhwn/41-us046-int-view-enrolled-respondents-for-a-survey-for-an-rumed)

ras_backstage/resources/reporting_units/get_reporting_unit.py
```diff
@@ -36,6 +38,9 @@ def get(ru_ref):
                                               if ce['id'] in case_collection_exercise_ids
                                               and parse_date(ce['scheduledStartDateTime']) < now]
 
+            respondents = respondents_for_survey.get(survey.get('id'))
+            survey['respondents'] = respondents if respondents else []
```
andrewmil: Instead of the if/else here you can change the above line to `respondents_for_survey.get(survey.get('id'), [])`

PR #58, opened 2018-02-21
Us029 add collection instrument link endpoint

andrewmil:
Added an endpoint which calls the collection instrument service to link a collection instrument to a collection exercise

This is the endpoint that will be called from response-operations-ui when an internal user selects an eQ questionnaire as the collection instrument for a collection exercise

[Trello](https://trello.com/c/QharabAR/42-us029-int-select-load-eq-collection-instruments-xxl)
[Confluence](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/248643743/US029+Select+EQ+Collection+Instruments+XXL)

### DIFF
--- a/ras_backstage/resources/collection_instrument/link_exercise.py
+++ b/ras_backstage/resources/collection_instrument/link_exercise.py
@@ -1,0 +1,25 @@
+import logging
+
+from flask import Response
+from flask_restplus import Resource
+from structlog import wrap_logger
+
+from ras_backstage import collection_instrument_api
+
+from ras_backstage.controllers.collection_instrument_controller import link_collection_instrument_to_exercise
+
+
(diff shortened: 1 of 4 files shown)


ras_backstage/resources/collection_instrument/link_exercise.py
```diff
@@ -0,0 +1,25 @@
+import logging
+
+from flask import Response
+from flask_restplus import Resource
+from structlog import wrap_logger
+
+from ras_backstage import collection_instrument_api
+
```
jcox-dev: Extra line can be removed here.

tej99:
[US029: INT Select/Load EQ Collection Instruments [XXL]](https://trello.com/c/QharabAR/42-us029-int-select-load-eq-collection-instruments-xxl)

PR #60, opened 2018-02-23
Integrate RAS backstage with UAA 

warrenbailey:
Integrate RAS backstage with the UAA token and provide secure messaging with the correct user information in the token


ras_backstage/__init__.py
```diff
@@ -15,6 +18,34 @@
...
 logger_initial_config(service_name='ras-backstage', log_level=app.config['LOGGING_LEVEL'])
+logger = logging.getLogger(__name__)
+
+
+def get_public_key():
+    headers = {
+        'Accept': 'application/json',
+    }
+
+    public_key_url = f'{app.config["UAA_SERVICE_URL"]}{"/token_key"}'
```
insacuri: 🎨Could be rewritten to:  ```f"{app.config["UAA_SERVICE_URL"]}/token_key"```

ras_backstage/__init__.py
```diff
@@ -15,6 +18,34 @@
...
+    response = requests.get(public_key_url, headers=headers)
+
+    try:
+        response.raise_for_status()
+    except HTTPError:
+        logger.exception("Error while retrieving public key")
+        raise
+
+    try:
+        key = response.json()['value']
```
insacuri: Response.json raises a valueError if it fails.  Should be caught here on the very slim chance that this might happen
http://docs.python-requests.org/en/latest/user/quickstart/#json-response-content

ras_backstage/authentication/token_decoder.py
```diff
@@ -0,0 +1,19 @@
+import jwt
+from flask import current_app
+
+
+def decode_access_token(access_token):
+    decoded_jwt = jwt.decode(
```
insacuri: Decode can throw a DecodeError, shouldn't this be handled?
https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jwt.py#L83

ras_backstage/controllers/uaa_controller.py
```diff
@@ -63,17 +38,10 @@ def sign_in(username, password):
...
-            algorithms=response.json().get('alg'),
-            verify=True,
-            key=get_public_key(),
-            audience='ras_backstage',
-            leeway=10,
-        )
-
         logger.debug('Successfully retrieved UAA token')
-        return decoded_jwt
+        token = response.json()
```
insacuri: Same as above, if the json is malformed or missing then you get a valueError, should be this be handled too?

test/resources/test_secure_messaging.py
```diff
@@ -11,12 +11,12 @@
 with open('test/test_data/secure_messaging/message.json') as json_data:
     message = json.load(json_data)
 url_get_messages = '{}{}{}'.format(app.config['RAS_SECURE_MESSAGING_SERVICE'],
-                                   'messages', '?label=INBOX&limit=1000')
+                                   'v2/messages', '?label=INBOX&limit=1000')
```
insacuri: 🎨 Could be rewritten to f"{app.config['RAS_SECURE_MESSAGING_SERVICE']}v2/messages?label=INBOX&limit=1000" .  Feel free to not bother though ^^, at some point there needs to be a pass over the whole repo to make this look tider

ras_backstage/__init__.py
warrenbailey: Good spot

warrenbailey:
@AndrewTorrance I'd argue that secure messaging should be changed to always use HS256. I'd doubt we'd change it via config and if it was need a simple code change will be easy enough. Less chance of configuration mismatch then.

ras_backstage/__init__.py
warrenbailey: Though any error in __init__.py will stop the application dead

ras_backstage/authentication/token_decoder.py
warrenbailey: yes

ras_backstage/controllers/uaa_controller.py
warrenbailey: I'll add an except block

ras_backstage/__init__.py
```diff
@@ -1,10 +1,11 @@
 import os
-
+import logging
```
JamesGardiner: logging should be before os and there should be a space before the flask import

ras_backstage/__init__.py
```diff
@@ -1,10 +1,11 @@
...
 from flask import Flask
 from flask_cors import CORS
 from flask_restplus import Api, Namespace
 
 from ras_backstage.logger_config import logger_initial_config
+from ras_backstage.authentication.uaa import request_uaa_public_key
```
JamesGardiner: alphabetical imports

ras_backstage/authentication/token_decoder.py
```diff
@@ -0,0 +1,33 @@
...
+        )
+        return decoded_jwt
+    except DecodeError:
+        logger.error(f"Unable to decode token {access_token} - confirm the UAA public key is correct")
+        abort(500)
+
+
+def get_user_id(access_token, verify=True):
+    decoded_jwt = decode_access_token(access_token, verify)
+    if decoded_jwt:
```
JamesGardiner: Will decoded_jwt ever evaluate to None? it looks like it will either succesfully decode or abort(500).

If not, this could be ```return decode_access_token(access_token, verify).get('user_id')```

In which case this function is obsolete.

ras_backstage/authentication/uaa.py
```diff
@@ -0,0 +1,35 @@
+import logging
+
+import requests
+from requests import HTTPError
+from json import JSONDecodeError
```
JamesGardiner: json is stdlib.

ras_backstage/authentication/uaa.py
```diff
@@ -0,0 +1,35 @@
...
+from flask import current_app
```
JamesGardiner: alphabetize

ras_backstage/authentication/uaa.py
```diff
@@ -0,0 +1,35 @@
...
+logger = logging.getLogger(__name__)
+
+
+def request_uaa_public_key(app):
+    headers = {
+        'Accept': 'application/json',
+    }
+
+    public_key_url = f'{app.config["UAA_SERVICE_URL"]}/token_key'
+    response = requests.get(public_key_url, headers=headers)
```
JamesGardiner: This needs to be handled correctly i.e. at least wrap it in a requests.RequestException handler

ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -1,9 +1,13 @@
 import json
 import logging
 
+
+import jwt
+from flask import current_app
```
JamesGardiner: PEP8 please

ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -13,8 +17,8 @@
 
 def get_messages_list(encoded_jwt, message_args):
     logger.debug('Retrieving messages list', label=message_args.get('label'))
-    url = '{}{}'.format(app.config['RAS_SECURE_MESSAGING_SERVICE'], 'messages')
-    headers = {"Authorization": encoded_jwt}
+    url = '{}{}'.format(app.config['RAS_SECURE_MESSAGING_SERVICE'], 'v2/messages')
```
JamesGardiner: f strings would be more concise.

ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -101,3 +105,21 @@ def update_draft(encoded_jwt, message_json):
...
     logger.debug('Draft updated successfully', message_id=message['msg_id'])
     return message
+
+
+def create_authorization_header(encoded_jwt):
+    if current_app.config.get('USE_UAA'):
+        logger.info(f'Received token {encoded_jwt}')
+        sm_token = convert_token(encoded_jwt)
+    else:
+        sm_token = jwt.encode({'party_id': 'BRES', 'role': 'internal'},
```
JamesGardiner: Does this need to catch any jwt encode errors?

ras_backstage/controllers/uaa_controller.py
```diff
@@ -1,9 +1,9 @@
 import logging
+from json import JSONDecodeError
 
-import jwt
-import requests
 from requests import HTTPError
 from structlog import wrap_logger
+from flask_restplus import abort
```
JamesGardiner: these need to be alphabetized

ras_backstage/authentication/token_decoder.py
warrenbailey: Good point, think it was from when I had decode_access_token return None on failure. I'll change

ras_backstage/authentication/uaa.py
warrenbailey: I specifically do this to annoy you ;)

ras_backstage/authentication/uaa.py
warrenbailey: Agree, should read the documentation ;)

ras_backstage/controllers/secure_messaging_controller.py
warrenbailey: Also all the secure messaging URLs need to be changed to v2 so I'll do the format change separately as part of that PR.

ras_backstage/controllers/secure_messaging_controller.py
JamesGardiner: Yeah I can only see the TypeError

config.py
```diff
@@ -109,16 +106,13 @@ class DevelopmentConfig(Config):
...
     BASIC_AUTH = (SECURITY_USER_NAME, SECURITY_USER_PASSWORD)
-    DJANGO_CLIENT_ID = os.getenv('DJANGO_CLIENT_ID', 'ons@ons.gov')
-    DJANGO_CLIENT_SECRET = os.getenv('DJANGO_CLIENT_SECRET', 'password')
-    DJANGO_BASIC_AUTH = (DJANGO_CLIENT_ID, DJANGO_CLIENT_SECRET)
-    JWT_SECRET = os.getenv('JWT_SECRET', 'testsecret')
     USERNAME = os.getenv('RAS_BACKSTAGE_USERNAME', 'user')
     PASSWORD = os.getenv('RAS_BACKSTAGE_PASSWORD', 'pass')
     USE_UAA = int(os.getenv('USE_UAA', '1'))
-    UAA_SERVICE_URL = os.getenv('UAA_SERVICE_URL', 'localhost')
+    UAA_SERVICE_URL = os.getenv('UAA_SERVICE_URL', 'http://localhost')
```
insacuri: This looks odd to me, uaa would be at localhost without a port?  Not a huge deal, just a thought I had while looking

JamesGardiner: defaults to 80?

warrenbailey: Added port

ras_backstage/authentication/token_decoder.py
```diff
@@ -0,0 +1,35 @@
...
+        )
+        return decoded_jwt
+    except jwt.DecodeError:
+        logger.error(f"Unable to decode token {access_token} - confirm the UAA public key is correct")
+        abort(500)
+
+
+def get_user_id(access_token, verify=True):
+    decoded_jwt = decode_access_token(access_token, verify)
+    user_id = decoded_jwt.get(USER_ID_CLAIM)
```
ajmaddaford: Why not just `decoded_jwt.get('user_id')`?

ras_backstage/authentication/token_decoder.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def decode_access_token(access_token, verify=True):
+
+    try:
+        decoded_jwt = jwt.decode(
+            access_token,
+            verify=verify,
+            algorithms=None,
+            key=get_uaa_public_key(),
```
ajmaddaford: I don't think we should be calling a method here to try and get the public key from the UAA. We should only need to get the public key in one place on startup and assume it's in the config after that. I also think the method should be provided with the key rather than get it from the config as this will be unit testable outside of a Flask context.

ras_backstage/authentication/token_decoder.py
```diff
@@ -0,0 +1,35 @@
...
+            audience='ras_backstage',
+            leeway=10,
+        )
+        return decoded_jwt
+    except jwt.DecodeError:
+        logger.error(f"Unable to decode token {access_token} - confirm the UAA public key is correct")
+        abort(500)
```
ajmaddaford: I don't think we should be returning a 500 error from here. Ideally all responses should be as close as possible to the route definition (e.g. in sign_in.py), it makes it much easier to see what the possible responses are for a given endpoint.

I think it would be clearer if we separated the decoding from getting the user_id. The route can then decide what to do with the decode error.

ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -101,3 +103,21 @@ def update_draft(encoded_jwt, message_json):
...
+                              app.config['RAS_SECURE_MESSAGING_JWT_SECRET'],
+                              algorithm='HS256')
+    return {"Authorization": sm_token}
+
+
+def convert_token(access_token):
+    user_id = token_decoder.get_user_id(access_token)
+
+    secret = app.config['RAS_SECURE_MESSAGING_JWT_SECRET']
+    return jwt.encode({'party_id': user_id, 'role': 'internal'}, secret, algorithm='HS256')
```
ajmaddaford: I think we could remove the convert_token method and remove some duplication to make this clearer, something like:
```
def _create_authorization_header(encoded_jwt):
    if current_app.config.get('USE_UAA'):
        logger.info(f'Received token {encoded_jwt}')
        party_id = token_decoder.get_user_id(access_token)
    else:
        party_id = 'BRES'	

	secret = app.config['RAS_SECURE_MESSAGING_JWT_SECRET']
	return jwt.encode({'party_id': party_id, 'role': 'internal'}, secret, algorithm='HS256')
```

ras_backstage/authentication/token_decoder.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def get_user_id(access_token, verify=True):
```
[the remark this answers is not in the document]
warrenbailey: Actually that's not entirely correct, the unit tests pass False to this method. The reasoning being the hard coded token in the tests has expired and therefore would fail verification.

I can remove the test and that will allow me to remove the default parameter?

ras_backstage/authentication/token_decoder.py
warrenbailey: Its the java in me, I naturally want to make strings constants

ras_backstage/authentication/token_decoder.py
warrenbailey: The get_uaa_public_key method first checks if the key is in app config, if not will go and retrieve it from the UAA and put it in the config. The config is also initialised on startup but if the UAA isn't available then this allows the application to start and retrieve the key later.

def get_uaa_public_key():
    if not current_app.config.get('UAA_PUBLIC_KEY'):
        current_app.config['UAA_PUBLIC_KEY'] = request_uaa_public_key(current_app)
    return current_app.config['UAA_PUBLIC_KEY']

JamesGardiner: I disagree that we should only be getting the key on startup. The key comes with an etag header  that allows UAA to know if the key is up to date. according to the docs, this expires on a configuration change, so could be relatively frequent compared to statically generated keys. There should be some way of refreshing that key without bringing down the service.

ras_backstage/authentication/token_decoder.py
warrenbailey: I'm not sure I agree with this comment. Abort 500 raises an InternalServerError which is handled by flasks internal error handling, so it'll forward to the standard error template. 

I could raise an error manually and deal with it in the error handler instead, but it achieves the same result.

ras_backstage/controllers/secure_messaging_controller.py
warrenbailey: Agreed, will change

ras_backstage/authentication/token_decoder.py
JamesGardiner: AFAIK for your first point - this could pollute the route definition quite a lot. Abort is seen as good practice in Flask and lets you reliably end a request and (as long as you have the correct error handlers in place) display the correct error message to the user.

warrenbailey: I'll remove the try except block in the token decoder and let it ripple up to the sign in page. It can abort from there. However that means @insacuri comment about error handling in the token decoder will have to be ignored.

ras_backstage/authentication/token_decoder.py
warrenbailey: That's a good point, however I would like to do that work separately. We're very unlikely to rotate the public key in the first release and if we do then an app restart is acceptable.

warrenbailey: https://trello.com/c/KgjFhxme/92-refresh-public-key

ajmaddaford: @warrenbailey My concern with this is it gives a false impression that the service is running, as it will start up, but the UAA may never start. The complexity of retrying to get a key could be done by whatever is orchestrating the deployment e.g. try to redeploy every minute for x minutes. One of the current frustrations in setting up all of the services is services that have started but don't work as they are dependent on other services that haven't. 

@JamesGardiner Currently once it's got a key, it puts it in the config and doesn't call the UAA again. Future work can look at how we might refresh that key, but we'd have to consider if we want to make a request to the UAA for a key on every decode.

ras_backstage/authentication/token_decoder.py
ajmaddaford: My comment was more about leaking Flask code into the token decoder. I was thinking of it more like we have a token decoder method that can raise an internal server error, which doesn't make sense to me.

ras_backstage/authentication/token_decoder.py
warrenbailey: A big problem is the UAA takes a lot longer to start than ras_backstage, so if we do that then the docker setup it will always fail to start. 

I agree about false impressions but in this scenario it will be obvious. The sign in will fail with an error message in the log

ras_backstage/authentication/token_decoder.py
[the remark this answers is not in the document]
ajmaddaford: We shouldn't be adding parameters that are only used by tests. I'm not sure we should be testing the decode of a UAA token (`decode_access_token` should probably be `_decode_access_token` and therefore not tested directly in tests). The test could test that given a decoded UAA token, we can get the user id out?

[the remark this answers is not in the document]
warrenbailey: I actually disagree on that, its not an uncommon thing to write code in such a way to make it unit testable, I used to do it with protected methods in Java all the time. 

However I have changed it, remove the flag, simplified the code using James original suggestion of removing the get_user_id method and doing a key look up on the returned dict instead. At that point the tests are redundant anyway as they were only testing PyJwt decode method then, so I've removed them.

warrenbailey:
Rebased to master

ras_backstage/authentication/token_decoder.py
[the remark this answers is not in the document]
ajmaddaford: Ok, we should discuss this as a team next week.

warrenbailey:
The build has passed, for some reason github hasn't updated. I'm retriggering the travis build

PR #64, opened 2018-02-26
Sm101 collection case

atrozzo:
integrated the collection case id in the response for the respondent to survey, when is present. The collection Id is resolved via cal to case service passing a a business party id. In case the "case_id" is empty and empty string will be passed over and warning logged. 

Code GetReportingUnit , line 77.

### DIFF
--- a/ras_backstage/resources/reporting_units/get_reporting_unit.py
+++ b/ras_backstage/resources/reporting_units/get_reporting_unit.py
@@ -75,7 +75,12 @@ def link_respondents_to_survey(respondents, survey):
 
         # Collect the collection case.
         json = get_cases_by_business_party_id(respondent.get('id'))
-        respondent['case_id'] = json[0].get('id')
+        try:
+            respondent['case_id'] = json[0].get('id')
+        except IndexError:
+            respondent['case_id'] = ''
+            logger.warning(f"no case id for respondent {respondent.get('id')}")
+
         for association in respondent.get('associations'):
             for enrolment in association.get('enrolments'):
                 respondent['enrolmentStatus'] = enrolment.get('enrolmentStatus')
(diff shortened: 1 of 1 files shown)


ras_backstage/resources/reporting_units/get_reporting_unit.py
```diff
@@ -71,6 +72,15 @@ def add_collection_exercise_details(collection_exercises, reporting_unit, cases)
 def link_respondents_to_survey(respondents, survey):
     survey['respondents'] = []
     for respondent in respondents:
+
+        # Collect the collection case.
+        json = get_cases_by_business_party_id(respondent.get('id'))
+        try:
+            respondent['case_id'] = json[0].get('id')
```
dcdarrell9: Can't we do this something like `respondent['case_id'] = json['id']`
And maybe name it something instead of `json`, maybe `cases` ?

dcdarrell9:
```
cases = case_controller.get_cases_by_business_party_id(respondent['id'])
        try:
            respondent['case_id'] = cases[0]['id']
```
This works for me, but I'm wondering about the cases[0] (0 is 'ACTIONABLE' and 1 'INACTIONABLE') I don't know what they're for.

ras_backstage/resources/reporting_units/get_reporting_unit.py
```diff
@@ -71,6 +72,15 @@ def add_collection_exercise_details(collection_exercises, reporting_unit, cases)
...
+        except IndexError:
+            respondent['case_id'] = ''
+            logger.warning(f"no case id for respondent {respondent.get('id')}")
```
dcdarrell9: and here is more preference for me  `respondent['id']`

ras_backstage/resources/reporting_units/get_reporting_unit.py
insacuri: This can also throw a KeyError if cases[0] is empty for whatever reason.  We need to be handling that exception too

insacuri (approved):

JamesGardiner:
Is this being closed?

AdamHawtin:
Closed until we can get the correct case ID.

PR #69, opened 2018-03-06
Sm124 get thread

atrozzo:
Created endpoint for retriving message conversations via thred id from secure message service.


config.py
```diff
@@ -30,6 +30,8 @@ class Config(object):
         RAS_SECURE_MESSAGING_SERVICE_PROTOCOL,
         RAS_SECURE_MESSAGING_SERVICE_HOST, RAS_SECURE_MESSAGING_SERVICE_PORT)
 
+    RAS_SECURE_MESSAGING_SERVICE_THREAD = f"{RAS_SECURE_MESSAGING_SERVICE}v2/threads/"
```
JamesGardiner: I don't think that this shouldn't be config; it doesn't change between environments, is a astable endpoint and isn't a secret that can't be put in the public domain.

ras_backstage/__init__.py
```diff
@@ -70,6 +70,7 @@
 from ras_backstage.resources.sign_in.sign_in import SignInV2  # NOQA # pylint: disable=wrong-import-position
 from ras_backstage.resources.survey.get_survey_list import GetSurveyList  # NOQA # pylint: disable=wrong-import-position
 from ras_backstage.resources.survey.get_survey_by_short_name import GetSurveyByShortName  # NOQA # pylint: disable=wrong-import-position
+from ras_backstage.resources.secure_messaging.get_thread import GetThread  # NOQA # pylint: disable=wrong-import-position
```
JamesGardiner: Can this go in with the other secure_message module imports.

ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -43,6 +43,27 @@ def get_message(encoded_jwt, message_id, is_draft):
...
+
+    headers = _create_authorization_header(encoded_jwt)
+    response = request_handler('GET', url, headers=headers)
+
+    if response.status_code != 200:
+        logger.error('Error retrieving the messages', thread_id=thread_id)
+        raise ApiError(url, response.status_code)
+
+    logger.debug('Successfully retrieved message', thread_id=thread_id)
+    return json.loads(response.text)
```
JamesGardiner:
```Python
return response.json()
```

also, you might want to guard against ValueErrors when decoding arbitrary JSON.

ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -43,6 +43,27 @@ def get_message(encoded_jwt, message_id, is_draft):
...
+    :param thread_id: is the id of the thread selected.
+    :return: a list of messages sent/received part of conversation thread
+    """
+
+    url = f"{app.config['RAS_SECURE_MESSAGING_SERVICE_THREAD']}{thread_id}"
+
+    headers = _create_authorization_header(encoded_jwt)
+    response = request_handler('GET', url, headers=headers)
+
+    if response.status_code != 200:
```
JamesGardiner: Do we need different actions based on error code ranges?

ras_backstage/resources/secure_messaging/get_thread.py
```diff
@@ -0,0 +1,31 @@
...
+class GetThread(Resource):
+    method_decorators = [get_jwt(request)]
+
+    @staticmethod
+    @secure_messaging_api.expect(parser)
+    @secure_messaging_api.header('Authorization',
+                                 'JWT to pass to secure messaging service', required=True)
+    def get(encoded_jwt, thread_id):
+
+        conversation_thread = secure_messaging_controller.get_thread_by_id(encoded_jwt, thread_id)
```
JamesGardiner: Does this need any error handling?

ras_backstage/__init__.py
dcdarrell9: Can we do from ras_backstage import resources? @JamesGardiner

ras_backstage/controllers/secure_messaging_controller.py
dcdarrell9: If it does have an impact might be worth adding tech debt for this as it's in multiple different places?

ras_backstage/__init__.py
JamesGardiner: To access this class you would then have to do `resources.get_thread.GetThread` which is overly verbose. Better to put it in the right import position.

ras_backstage/controllers/secure_messaging_controller.py
JamesGardiner: The impact would be that the whole process would exit with a non-zero code when the ValueError is raised (I do not think this is caught elsewhere). If it is in multiple places, then we need to evaluate the risk and decide whether to guard against this happening (which it eventually will).

JamesGardiner: If we do decide to fix it elsewhere, that doesn't mean allowing this to get in to the codebase is ok.

config.py
atrozzo: Still not sure why properties need to be someothing that change over enviroments. In properties you place what actually never change but might be used in different place on the coode. Search/replace string is an infamous bad practice. For me if you hav ea String that is repted more then once in ifferent code place should be placed somewhere common. How a conf/property file can be worse then search string in the code ?

ras_backstage/__init__.py
atrozzo: done it.

ras_backstage/controllers/secure_messaging_controller.py
atrozzo: done it.

ras_backstage/controllers/secure_messaging_controller.py
atrozzo: can be 200 or 500.  If you se eall teh other roots hav ethe same http status check.

ras_backstage/resources/secure_messaging/get_thread.py
atrozzo: Add some defensive code.

config.py
warrenbailey: I'm not sure I understand your comment there Angello but I agree with James this doesn't need to be in the flask config the endpoint shouldn't change.

ras_backstage/controllers/secure_messaging_controller.py
warrenbailey: Agree, they all do this and we should have a review of all error handling in backstage but for now I think this should stay as it is for consistency

ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -43,6 +44,30 @@ def get_message(encoded_jwt, message_id, is_draft):
...
+        logger.error('Error retrieving the messages', thread_id=thread_id)
+        raise ApiError(url, response.status_code)
+
+    try:
+        return json.loads(response.text)
```
warrenbailey: return response.json()

ras_backstage/resources/secure_messaging/get_thread.py
```diff
@@ -0,0 +1,38 @@
...
+
+        try:
+            return make_response(jsonify(secure_messaging_controller.get_thread_by_id(encoded_jwt, thread_id)), 200)
+        except ApiError:
+            err_msg = f"Failed to retrieve Thread data with thread id: {thread_id}"
+
+        except JSONDecodeError:
+            err_msg = f"Secure message returned a malformed Json for id: {thread_id}"
+
+        logger.exception(err_msg)
```
warrenbailey: This needs to be in the exception block to log the exception

atrozzo: changed in error.

ras_backstage/controllers/secure_messaging_controller.py
atrozzo: not convinced to do the right thing changing it. I do not see the perfomance and memery outome of this.

ras_backstage/controllers/secure_messaging_controller.py
atrozzo: changed

ras_backstage/controllers/secure_messaging_controller.py
atrozzo: changed

config.py
atrozzo: changed

PR #70, opened 2018-03-06
SM111 Add get threads endpoint

AdamHawtin:
For story SM111 response-operations-ui needs to get messages threads for it's inbox view. This PR adds a 'threads' endpoint to the backstage api which gets threads from the secure message threads endpoint.

Trello Card: 
[SM111 Int View All Inbox Messages From Threads](https://trello.com/c/gfaiP90r/60-sm111-int-view-all-inbox-messages-from-threads-13)


ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -43,6 +43,21 @@ def get_message(encoded_jwt, message_id, is_draft):
     return json.loads(response.text)
 
 
+def get_threads_list(encoded_jwt, message_args):
+    logger.debug('Retrieving threads list')
+    url = f"{app.config['RAS_SECURE_MESSAGING_SERVICE']}threads"
```
JamesGardiner: I don't think this should be in config

ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -43,6 +43,21 @@ def get_message(encoded_jwt, message_id, is_draft):
...
+    headers = _create_authorization_header(encoded_jwt)
+
+    response = request_handler('GET', url, headers=headers, params=message_args)
+
+    if response.status_code != 200:
```
JamesGardiner: Does this need to handle different error ranges for client vs server errors?

ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -43,6 +43,21 @@ def get_message(encoded_jwt, message_id, is_draft):
...
+        logger.error('Error retrieving the threads list')
+        raise ApiError(url, response.status_code)
+
+    logger.debug('Successfully retrieved the threads list')
+    return json.loads(response.text)
```
JamesGardiner:
```Python
return response.json()
```

ras_backstage/resources/secure_messaging/get_thread_list.py
```diff
@@ -0,0 +1,38 @@
...
+                                 'JWT to pass to secure messaging service', required=True)
+    def get(encoded_jwt):
+        message_args = {
+            'limit': request.args.get('limit', 1000),
+            'survey': request.args.get('survey')
+        }
+
+        logger.info('Retrieving threads list')
+
+        messages = secure_messaging_controller.get_threads_list(encoded_jwt, message_args)
```
JamesGardiner: Does this need any error handling?

ras_backstage/controllers/secure_messaging_controller.py
JamesGardiner: Also - be careful with value errors when decoding arbitrary JSON

ras_backstage/controllers/secure_messaging_controller.py
warrenbailey: I think we need to have a review of the error handling in backstage because they all do this.

ras_backstage/resources/secure_messaging/get_thread_list.py
```diff
@@ -0,0 +1,38 @@
...
+    @secure_messaging_api.expect(parser)
+    @secure_messaging_api.header('Authorization',
+                                 'JWT to pass to secure messaging service', required=True)
+    def get(encoded_jwt):
+        message_args = {
+            'limit': request.args.get('limit', 1000),
+            'survey': request.args.get('survey')
+        }
+
+        logger.info('Retrieving threads list')
```
warrenbailey: Probably should be debug level

ras_backstage/resources/secure_messaging/get_thread_list.py
warrenbailey: The API error in secure messaging will bubble up to the error handler and then returned as JSON, so as it stands not much point doing anything else here at the minute.

ras_backstage/resources/secure_messaging/get_thread_list.py
```diff
@@ -0,0 +1,38 @@
...
+
+        messages = secure_messaging_controller.get_threads_list(encoded_jwt, message_args)
+
+        logger.info('Successfully retrieved threads list')
```
warrenbailey: Again debug level is more appropriate

ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -45,6 +44,26 @@ def get_message(encoded_jwt, message_id, is_draft):
...
+        logger.error('Error retrieving the threads list')
+        raise ApiError(url, response.status_code)
+
+    logger.debug('Successfully retrieved the threads list')
+
+    try:
+        return response.json()
+    except ValueError:
+        logger.error('Failed to decode response json from get threads')
+        raise ApiError(url=url)
```
dcdarrell9: Do you need this? Can you just have url?

PR #74, opened 2018-03-08
Add resend verification email endpoint

jwgwalton:


ras_backstage/controllers/party_controller.py
```diff
@@ -62,3 +62,15 @@ def get_businesses_by_search(query):
 
     logger.debug('Successfully retrieved businesses by search query')
     return json.loads(response.text)
+
+
+def resend_verification_email(party_id):
+    logger.debug('Resending verification email')
```
andrewmil: We should log out the `party_id` in the log messages for this function

jwgwalton: in the resource it's 

        logger.info('Attempting to resend verification email', party_id=party_id)
        party_controller.resend_verification_email(party_id)
        logger.info('Successfully resent verification email', party_id=party_id)

Wasn't sure if that'd add anymore, especially as it's a debug.

andrewmil: Yeah that's true, it's just the standard pattern we've run with so far. In the future we might use this controller from another resource which might be more complex and this kind of detail is good for those cases. For now it's fine though I'm sure

jwgwalton: Nah that sounds like a good reason to do it, I'll change it now

PR #84, opened 2018-03-20
Add new values for view.

rjweeks:


ras_backstage/resources/survey/get_survey_by_short_name.py
```diff
@@ -24,8 +24,21 @@ def get(short_name):
         survey['shortName'] = format_short_name(survey['shortName'])
 
         ce_list = collection_exercise_controller.get_collection_exercises_by_survey(survey['id'])
-
+        GetSurveyByShortName.add_events(ce_list)
```
andrewmil: `self.add_events` is a more standard way to write this

tej99: Why is self. a more standard way to use static methods?

jcox-dev: Because it's being called from within a bound method by an instance of the class with that static method. Both implementations do the same thing but `self` tends to read better in this context (as it provides a prompt for the reader that may otherwise go to check where this class was imported from, etc.)

rjweeks: This method is also static sp you can't use self in this context.

jcox-dev: Hah! Missed that entirely :)

andrewmil:
Looks good, we do need tests for the new endpoint though

PR #86, opened 2018-03-21
Update the update message label endpoint to V2

AdamHawtin:
Changes the secure message end point that is used to update the message labels to the v2 version.
Also renames the class as the previous name was misleading, it could do more than just remove the unread label on a message.

Also minor unrelated linting fixes necessary to run `make test` successfully.

**How to test:**
There is a branch of response-operations-ui that uses this endpoint:
[Response Operations UI: SM114/distinguish unread messages](https://github.com/ONSdigital/response-operations-ui/tree/sm114/distinguish-unread-messages)

And try opening an unread message in the secure message inbox.

### DIFF
--- a/ras_backstage/controllers/secure_messaging_controller.py
+++ b/ras_backstage/controllers/secure_messaging_controller.py
@@ -89,7 +89,7 @@ def get_thread_by_id(encoded_jwt, thread_id):
 
 def update_label(encoded_jwt, message_id, label, action):
     logger.debug('Updating label', message_id=message_id, label=label, action=action)
-    url = f"{app.config['RAS_SECURE_MESSAGING_SERVICE']}message/{message_id}/modify"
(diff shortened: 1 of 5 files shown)


ras_backstage/controllers/secure_messaging_controller.py
```diff
@@ -89,7 +89,7 @@ def get_thread_by_id(encoded_jwt, thread_id):
 
 def update_label(encoded_jwt, message_id, label, action):
     logger.debug('Updating label', message_id=message_id, label=label, action=action)
-    url = f"{app.config['RAS_SECURE_MESSAGING_SERVICE']}message/{message_id}/modify"
+    url = f"{app.config['RAS_SECURE_MESSAGING_SERVICE']}v2/messages/modify/{message_id}"
```
JamesGardiner: This really should be a PUT on the message ID.

PR #95, opened 2018-04-12
Removed all secure message functionality

insacuri:
Needs to be merged in at the same time as https://github.com/ONSdigital/response-operations-ui/pull/122.

IMPORTANT: This removes configuration from the service.  This needs to be updated in jenkins


Jenkinsfile
```diff
@@ -35,9 +35,6 @@ pipeline {
                 sh "cf set-env ras-backstage-service-dev RAS_OAUTH_SERVICE_HOST ras-django-dev.${env.CF_DOMAIN}"
                 sh "cf set-env ras-backstage-service-dev RAS_OAUTH_SERVICE_PORT 80"
 
-                sh "cf set-env ras-backstage-service-dev RAS_SECURE_MESSAGING_SERVICE_HOST ras-secure-messaging-dev.${env.CF_DOMAIN}"
```
dcdarrell9: I don't think these Jenkinsfiles are getting used at the moment, might be worth asking Ben J, pretty sure we're not going to be keeping them

dcdarrell9: If I remember right it's out of date anyway, I think a lot of the repos were going to be using these but we went a different way... Everything from this pr is up to date so not an issue for me  👍

PR #96, opened 2018-04-12
Us022 disable enrolment status

andrewmil:
Added endpoint for updating a respondents enrolment status for a survey.

This brought up a bug in returning the enrolment status to the UI on the reporting unit page. We were editing a single dictionary a large number of times to create the response and the status was being overridden when there was more than one survey. Things have been made more functional with list comprehensions to prevent this.

Works with other PR's:
[response-operations-ui](https://github.com/ONSdigital/response-operations-ui/pull/123)
[ras-party]()
[rm-case]()
[rm-casesvc-api]()

[Confluence](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/248545529/US022+Disable+respondent+enrolment+to+a+survey)
[Trello](https://trello.com/c/1ez85L0y/51-us022-int-disable-respondents-enrolment-to-a-survey-l)

### DIFF
--- a/ras_backstage/resources/reporting_units/get_reporting_unit.py
+++ b/ras_backstage/resources/reporting_units/get_reporting_unit.py
@@ -44,14 +44,24 @@ class GetReportingUnit(Resource):
         # Get all respondents for the given ru
         respondents = [party_controller.get_party_by_respondent_id(respondent['partyId'])
                        for respondent in reporting_unit.get('associations')]
+        respondents_in_business = [respondent
+                                   for respondent in respondents
+                                   if ru_ref in
+                                   [association['sampleUnitRef']
+                                    for association in respondent['associations']]
+                                   ]
 
         # Link collection exercises and respondents to surveys
         cases = case_controller.get_cases_by_business_party_id(reporting_unit['id'])
         for survey in surveys:
+            respondents_in_survey = [respondent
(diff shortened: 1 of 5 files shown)


gemmai95:
[US022: INT Disable respondents enrolment to a survey [L]](https://trello.com/c/1ez85L0y/51-us022-int-disable-respondents-enrolment-to-a-survey-l)

ras_backstage/resources/reporting_units/get_reporting_unit.py
```diff
@@ -62,6 +66,23 @@ def get(ru_ref):
         return make_response(jsonify(response_json), 200)
 
 
+def survey_ids_for_respondent(respondent, ru_ref):
+    return [enrolment.get('surveyId')
```
jwgwalton: I would split this into two list comprehensions
respondent_enrolments = [association.get('enrolments') for association in respondent.get('associations') if association['sampleUnitRef'] == ru_ref][0]]

return [enrolement.get('surveyId) for enrolment in respondent_enrolments]

The nested list comprehensions are a bit complicated.

andrewmil: :+1: I agree with this
There are one or two places where we still have nested list comprehensions but usually they're pretty tightly bound together

PR #97, opened 2018-04-13
US106 Edit survey details

emilioward:
Works alongside:
https://github.com/ONSdigital/rm-survey-service/pull/31 
https://github.com/ONSdigital/response-operations-ui/pull/124

This PR allows for an internal user to edit a survey `short_name` & `long_name`


emilioward:
[US106: INT User to Edit/Amend Survey (L)](https://trello.com/c/BGNKc7cM/63-us106-int-user-to-edit-amend-survey-l)

ras_backstage/controllers/survey_controller.py
```diff
@@ -83,3 +83,24 @@ def get_survey_ci_classifier(survey_id):
...
+        "ShortName": updated_survey_details['short_name'],
+        "LongName": updated_survey_details['long_name']
+    }
+
+    response = request_handler('PUT', url, auth=app.config['BASIC_AUTH'], json=payload)
+
+    if response.status_code == 404:
+        logger.error('Error retrieving survey details', surveyRef=surveyRef)
+        raise ApiError(url, response.status_code)
+    if response.status_code not in (200, 201, 202):
```
andrewmil: A cleaner way of writing this is `if not response.ok`

ras_backstage/controllers/survey_controller.py
```diff
@@ -83,3 +83,24 @@ def get_survey_ci_classifier(survey_id):
                  ci_selector=ci_selector['id'])
 
     return response.json()
+
+
+def update_survey_details(survey_ref, updated_survey_details):
+    surveyRef = survey_ref
```
andrewmil: ?

ras_backstage/resources/survey/update_survey_details.py
```diff
@@ -0,0 +1,30 @@
...
+})
+
+
+@survey_api.route('/edit-survey-details/<survey_ref>')
+class UpdateSurveyDetails(Resource):
+    @staticmethod
+    @survey_api.expect(survey_details, validate=True)
+    def put(survey_ref):
+
+        logger.info('Retrieving updated survey details', survey_ref=survey_ref)
```
andrewmil: This should probably be `Updating survey details`

ras_backstage/resources/survey/update_survey_details.py
```diff
@@ -0,0 +1,30 @@
...
+
+survey_details = survey_api.model('SurveyDetails', {
+    'short_name': fields.String(required=True),
+    'long_name': fields.String(required=True)
+})
+
+
+@survey_api.route('/edit-survey-details/<survey_ref>')
+class UpdateSurveyDetails(Resource):
+    @staticmethod
```
andrewmil: There should be a space between the class and the function

test/resources/test_survey.py
```diff
@@ -6,13 +6,16 @@
 
 from ras_backstage import app
 
+surveyRef = '023'
+survey_ref = '023'
```
andrewmil: We probably don't need both of these?

test/resources/test_survey.py
```diff
@@ -165,3 +171,43 @@ def test_get_survey_by_short_name_ce_empty(self, mock_request):
...
+        mock_request.put(url_update_survey_details, status_code=404)
+        url = f'backstage-api/v1/survey/edit-survey-details/{survey_ref}'
+        response = self.app.put(url, headers=self.headers, data=json.dumps({
+            "long_name": 'Survey Test Name',
+            "short_name": 'QBZ'
+        }))
+        self.assertEqual(response.status_code, 404)
+
+    @requests_mock.mock()
+    def test_edit_survey_details_no_data(self, mock_request):
```
andrewmil: Are we trying to test the restplus validation here? if so we don't need the mock and the response status we'd want to check for is a 400

ras_backstage/controllers/survey_controller.py
```diff
@@ -83,3 +83,23 @@ def get_survey_ci_classifier(survey_id):
...
+    url = f'{app.config["RM_SURVEY_SERVICE"]}surveys/ref/{survey_ref}'
+    payload = {
+        "ShortName": updated_survey_details['short_name'],
+        "LongName": updated_survey_details['long_name']
+    }
+
+    response = request_handler('PUT', url, auth=app.config['BASIC_AUTH'], json=payload)
+
+    if response.status_code == 404:
+        logger.error('Error retrieving survey details', survey_ref=survey_ref)
```
tej99: Should this be logged at error level?

andrewmil: We throw an ApiError right after so probably yes? If we get this response something must have gone wrong

tej99: If that's the case we should have error level logs in a lot more places. But I don't think a 404 should result in an error log because it gives us a splunk alert for something that does not require an action. warning level maybe better?

